PR #11, opened 2018-09-27
Add VoucherUnlinkCommandTest

t2d:
Replace test from sf3 with LiipFunctionalTestBundle


t2d:
phpunit-coverage in master: 0m1.178s
phpunit-coverage in this branch: 0m1.731s

but we get 15% more code coverage

PR #22, opened 2018-10-05
Add support for alias management to frontend

doobry-systemli:
Features:
* Use soft-delete for aliases in order to prevent impersonation
* List own aliases in user frontend
* Add new form to allow deletion of own aliases
* Allow to create new random aliases in user frontend
* Limit aliases per user to 20. Don't allow creation if limit is reached

Internal changes:
* Add new attributes `deleted` and `user_id` to entity `Alias`
* Make entity Alias implement SoftDeletableInterface
* Create indexes for relations {user_id,source,destination}<->deleted on
  entity `Alias`
* New classes `AliasFactory`, `AliasCreator` and `AliasHandler`
* Use `AliasFactory` to create aliases in fixtures
* Add validation constraints for entity `Alias`


src/Creator/AliasCreator.php
```diff
@@ -0,0 +1,48 @@
...
+            $alias = $this->validateAndUpdate($alias);
+        }
+
+        $this->validate($alias, ['Default', 'unique']);
+        $this->save($alias);
+
+        return $alias;
+    }
+
+    private function validateAndUpdate(Alias $alias)
```
0x46616c6b: Missing Docblick and return value

src/Creator/AliasCreator.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        try {
+            $this->validate($alias, ['unique']);
+        } catch (ValidationException $e) {
+            $localPart = RandomStringGenerator::generate(24, false);
+            $domain = $alias->getDomain();
+            $alias->setSource($localPart."@".$domain->getName());
+            $this->validateAndUpdate($alias);
+        }
+        return $alias;
```
0x46616c6b: Missing new line in front of return

src/Factory/AliasFactory.php
```diff
@@ -0,0 +1,34 @@
...
+        $domain = $user->getDomain();
+        $alias = new Alias();
+        $alias->setUser($user);
+        $alias->setDomain($domain);
+        $alias->setDestination($user->getEmail());
+        if (null === $localPart) {
+            $localPart = RandomStringGenerator::generate(24, false);
+        }
+
+        $alias->setSource($localPart."@".$domain);
```
0x46616c6b: `$domain` must be a string and you give a `Domain` object.

0x46616c6b: It would be more explicit when you call `getName` oder `__toString` method.

src/Handler/AliasHandler.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->creator = $creator;
+    }
+
+    /**
+     * @param User $user
+     * @return bool
+     */
+    public function checkAliasLimit(User $user): bool
+    {
+        $aliases = $this->repository->findByUser($user);
```
0x46616c6b: I would use a second parameter with an array of aliases, which we already fetched in StartController. So you reduce the database queries.

src/Handler/AliasHandler.php
```diff
@@ -0,0 +1,65 @@
...
+        return (count($aliases) < self::ALIAS_LIMIT) ? true : false;
+    }
+
+    /**
+     * @param User $user
+     * @param null|string $localPart
+     * @return bool
+     * @throws ValidationException
+     */
+    public function create(User $user, ?string $localPart): bool
```
0x46616c6b: I would return the alias if they created one or null.

Issue #23, opened 2018-10-08
Don't list redeemed vouchers


doobry-systemli:
I don't see a good reason to list redeemed vouchers in the frontend. If we would not list them, we could change the list order back to normal (list oldest ones first).

If we do this, we have to take care that the displayed message in case of no left vouchers is different than the one displayed if no vouchers are created yet.

What do you think?

t2d:
Yes, don’t list them. But still some users could have many unredeemed vouchers

doobry-systemli:
> Yes, don’t list them.

Ok will do so.

> But still some users could have many unredeemed vouchers

That's why I would keep the "create voucher" button on top.

PR #30, opened 2018-10-12
VoucherHandler: Add support to filter out redeemed vouchers (Fixes: #23)

doobry-systemli:
* Only list active vouchers in user frontend

### DIFF
--- a/src/Handler/VoucherHandler.php
+++ b/src/Handler/VoucherHandler.php
@@ -40,10 +40,11 @@ class VoucherHandler
 
     /**
      * @param User $user
+     * @param bool $redeemed
      *
      * @return Voucher[]
      */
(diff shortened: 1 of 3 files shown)


src/Handler/VoucherHandler.php
```diff
@@ -40,31 +40,37 @@ public function __construct(ObjectManager $manager, VoucherCreator $creator)
...
+                    }
                 }
             }
         }
 
-        return $vouchers;
+        if (true === $redeemed) {
+            return $vouchers;
+        } else {
+            $vouchersActive = array_filter($vouchers, function($k) {
```
0x46616c6b: You should give $k an Type (`Voucher`) and name it also `$voucher`

Issue #33, opened 2018-10-16
`checkpassword` console command (for dovecot authentication)


doobry-systemli:
Dovecot documentation for `checkpassword`: https://wiki2.dovecot.org/AuthDatabase/CheckPassword

The script should execute `$user->updateLastLogin()` in order to update the `lastLogin` timestamp.

t2d:
The current config (according to install instructions) produce the following errors:
```
Nov  2 13:37:06 mail dovecot: auth: Error: prefetch(user@example.org): userdb lookup not possible with only userdb prefetch
Nov  2 13:37:06 mail dovecot: lda(user@example.org): Error: user philipp.knopp@systemli.org: Auth USER lookup failed
Nov  2 13:37:06 mail dovecot: lda: Fatal: Internal error occurred. Refer to server log for more information.
```
The doc on [Prefetch](https://wiki2.dovecot.org/UserDatabase/Prefetch) has more information: 
> If you're using LDA or LMTP, you still need a valid userdb which can be used to locate the users. You can do this by adding a normal SQL/LDAP userdb after the userdb prefetch. The order of definitions is significant. See below for examples. 

I think, we might need another command for `userdb`.

t2d:
Let's just expect delivery via `lmtp`. This is generally more usable.

t2d:
LMTP doesn't help. It still needs a specific `userdb` lookup:
```
Nov  8 00:14:04 imap dovecot: lmtp(8131): Connect from 0.0.0.0
Nov  8 00:14:04 imap dovecot: auth: Error: prefetch(user@systemli.org,0.0.0.0): userdb lookup not possible with only userdb prefetch
```
Otherwise dovecot doesn't know if a user exists. 

If I try to re-use it as `userdb`
```
passdb {
  driver = checkpassword
  args = /path/to/user-management/bin/console usrmgmt:users:checkpassword
}

userdb {
  driver = checkpassword
  args = /path/to/user-management/bin/console usrmgmt:users:checkpassword
}
```

there is another error:
```
Nov  7 23:51:06 imap dovecot: lmtp(1749): Connect from 0.0.0.0
Nov  7 23:51:07 imap dovecot: auth: Error: [2018-11-07 23:51:07] console.ERROR: Error thrown while ru
nning command "{command}". Message: "{message}" {"exception":"[object] (Symfony\\Component\\Console\\
Exception\\InvalidArgumentException(code: 0): Invalid input format. See https://cr.yp.to/checkpwd/int
[...]

PR #36, opened 2018-10-19
WIP: Add checkpassword console command for user authentication (Fixes: #33)

doobry-systemli:
* Implemented according to the following specifications:
  - qmail: https://cr.yp.to/checkpwd/interface.html
  - dovecot: https://wiki2.dovecot.org/AuthDatabase/CheckPassword

### DIFF
--- a/src/Command/CheckPasswordCommand.php
+++ b/src/Command/CheckPasswordCommand.php
@@ -1,0 +1,167 @@
+<?php
+
+namespace App\Command;
+
+use App\Handler\UserAuthenticationHandler;
+use App\Repository\UserRepository;
+use Doctrine\Common\Persistence\ObjectManager;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Exception\InvalidArgumentException;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Process\Exception\ExceptionInterface as ProcessException;
+use Symfony\Component\Process\Process;
+
+/**
+ * @author doobry <doobry@systemli.org>
+ */
+class CheckPasswordCommand extends ContainerAwareCommand
+{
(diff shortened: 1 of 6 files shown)


src/Helper/FileDescriptorReader.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        $fileHandler = fopen($inputFd, 'r');
+
+        return stream_get_contents($fileHandler);
+    }
+
+    /**
+     * @return bool|string
+     */
+    public function readFd3()
```
0x46616c6b: The methods in this class could be static

src/Command/CheckPasswordCommand.php
```diff
@@ -42,11 +48,13 @@ class CheckPasswordCommand extends ContainerAwareCommand
      * CheckPasswordCommand constructor.
      *
      * @param ObjectManager             $manager
+     * @param FileDescriptorReader      $reader
      * @param UserAuthenticationHandler $handler
      */
-    public function __construct(ObjectManager $manager, UserAuthenticationHandler $handler)
+    public function __construct(ObjectManager $manager, FileDescriptorReader $reader, UserAuthenticationHandler $handler)
```
0x46616c6b: If the methods in FileDescriptorReader are static, we do not need to inject it in the contructor.

doobry-systemli:
I now outsourced `FileDescriptorReader` and wrote a unit test for `CheckPasswordCommand`. Test coverage is 100% for the Command.

src/Helper/FileDescriptorReader.php
doobry-systemli: Mh, but apparently static methods cannot be mocked in unit tests.

src/Command/CheckPasswordCommand.php
doobry-systemli: If we don't inject `FileDescriptorReader`, then we cannot mock it in the unit tests, right? And that's the whole purpose of `FileDescriptorReader`, to mock it in unit tests in order to make `CheckPasswordCommand` testable. Or is there another way that I don't know of?

PR #41, opened 2018-10-23
Add form for custom aliases

t2d:
Right now, I expose the validator error message. On the good side, it shows why an alias couldn't be created. On the bad side:

1) Allows to see if an email address already exists
2) Error is always shown in English

Number 1) is a bigger problem than in registration because users have the possibility to create and delete an infinite amount of aliases. Which makes enumeration easy. Do you have an idea how to handle this?

### DIFF
--- a/default_translations/de/messages.de.yml
+++ b/default_translations/de/messages.de.yml
@@ -10,9 +10,9 @@ index:
   voucher-headline: Deine Einladungscodes
   voucher-limit: Du hast keine Einladungscodes mehr verfügbar.
   voucher-disable: Du bekommst am %date% Uhr drei Einladungscodes gutgeschrieben. Dies dient der Vermeidung von Massenregistrierungen und Spam.
-  alias-headline: Deine Aliase
-  alias-limit: Du hast das Limit  von %alias_limit% erlaubten %alias_type% Aliases erreicht. Wenn du alte Aliases löschst, kannst du wieder neue erstellen.
-  alias-empty: Du hast noch keine Aliase definiert.
+  alias-headline: Deine %alias_type% Aliase
+  alias-limit: Du hast das Limit von %alias_limit% erlaubten %alias_type% Aliases erreicht. Wenn du alte Aliases löschst, kannst du wieder neue erstellen.
+  alias-empty: Du hast noch keine %alias_type% Aliase definiert.
   delete-headline: Account löschen
   delete-description: >
     Du kannst deinen Account bei %project_name% löschen. Dadurch werden deine E-Mails unwiderruflich gelöscht. Eine <strong>spätere Wiederherstellung ist nicht möglich.</strong>
--- a/default_translations/de/messages.de.yml
+++ b/default_translations/de/messages.de.yml
@@ -11,7 +11,7 @@ index:
   voucher-limit: Du hast keine Einladungscodes mehr verfügbar.
   voucher-disable: Du bekommst am %date% Uhr drei Einladungscodes gutgeschrieben. Dies dient der Vermeidung von Massenregistrierungen und Spam.
   alias-headline: Deine Aliase
-  alias-limit: Du hast das Limit  von %alias_limit% erlaubten Aliases erreicht. Wenn du alte Aliases löschst, kannst du wieder neue erstellen.
+  alias-limit: Du hast das Limit  von %alias_limit% erlaubten %alias_type% Aliases erreicht. Wenn du alte Aliases löschst, kannst du wieder neue erstellen.
   alias-empty: Du hast noch keine Aliase definiert.
   delete-headline: Account löschen
   delete-description: >
@@ -125,3 +125,6 @@ error:
 
 # internal translations
 Bad credentials.: Fehlerhafte Zugangsdaten
+
+custom: spezifischen
+random: zufälligen
--- a/default_translations/de/messages.de.yml
+++ b/default_translations/de/messages.de.yml
@@ -36,12 +36,14 @@ form:
   voucher-redeemed-on: Eingelöst am
   oclock-by: Uhr von
   actual-password: Aktuelles Passwort
+  new-custom-alias: Neues spezifisches Alias
   new-password: Neues Passwort
   new-password_confirmation: Neues Passwort bestätigen
   change-password: Passwort ändern
   delete-account: Account löschen
   delete-password: Passwort
   create-voucher: Erstelle Einladungscode
+  create-custom-alias: Erstelle spezifisches Alias
   create-random-alias: Generiere Zufalls-Alias
   delete-alias: Alias löschen
 
@@ -108,6 +110,7 @@ flashes:
   password-change-successful: Dein neues Passwort ist ab sofort aktiv!
   registration-successful: Du hast dich erfolgreich registriert!
   voucher-creation-successful: Du hast einen neuen Einladungscode erstellt!
+  custom-alias-creation-successful: Dein neues Alias wurde erstellt!
   random-alias-creation-successful: Dir wurde ein neues Zufalls-Alias erstellt!
   alias-deletion-successful: Dein Alias wurde erfolgreich gelöscht!
 
--- a/default_translations/de/messages.de.yml
+++ b/default_translations/de/messages.de.yml
@@ -36,14 +36,14 @@ form:
   voucher-redeemed-on: Eingelöst am
   oclock-by: Uhr von
   actual-password: Aktuelles Passwort
-  new-custom-alias: Neues spezifische Alias-Adresse
+  new-custom-alias: Neue Alias-Adresse
   new-password: Neues Passwort
   new-password_confirmation: Neues Passwort bestätigen
   change-password: Passwort ändern
   delete-account: Account löschen
   delete-password: Passwort
   create-voucher: Erstelle Einladungscode
-  create-custom-alias: Erstelle spezifische Alias-Adresse
+  create-custom-alias: Erstelle Alias-Adresse
   create-random-alias: Generiere zufällige Alias-Adresse
   delete-alias: Alias-Adresse löschen
 
@@ -125,5 +125,5 @@ error:
 # internal translations
 Bad credentials.: Fehlerhafte Zugangsdaten
 
-custom: spezifischen
+custom: gewählten
 random: zufälligen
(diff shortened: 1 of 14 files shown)


t2d:
I would like some feedback on this please. I'm not really happy about the distinction between custom and random aliases. It produced some duplicated code. But I didn't come up with a better idea.

src/Entity/Alias.php
```diff
@@ -26,12 +26,23 @@ class Alias implements SoftDeletableInterface
...
+    /**
+     * @var bool
+     */
+    protected $random;
+
+    /**
+     * @var int
+     */
+    protected $type;
+
```
doobry-systemli: `$type` seems to be a leftover?

src/Entity/Alias.php
```diff
@@ -70,4 +81,19 @@ public function setEmptyUser()
     {
         $this->user = null;
     }
+
+    public function getType()
+    {
+        return $this->type;
+    }
```
doobry-systemli: `getType()` seems to be a leftover?

src/Factory/AliasFactory.php
```diff
@@ -25,10 +27,11 @@ public static function create(User $user, ?string $localPart): Alias
         $alias->setDomain($domain);
         $alias->setDestination($user->getEmail());
         if (null === $localPart) {
-            $localPart = RandomStringGenerator::generate(24, false);
+            $localPart = RandomStringGenerator::generate(self::RANDOM_ALIAS_LENGTH, false);
+            $alias->setRandom();
         }
 
-        $alias->setSource($localPart.'@'.$domain->getName());
+        $alias->setSource(strtolower($localPart).'@'.$domain->getName());
```
doobry-systemli: I think it would be better to do `strtolower()` in the `AliasHandler` and add a check for lowercase strings to `EmailValidator` (if it's not there yet). That way, `AliasFactory` stays stupid and doesn't have to care about string manipulation.

src/Repository/AliasRepository.php
```diff
@@ -39,4 +39,14 @@ public function findByUser(User $user)
     {
         return $this->findBy(['user' => $user]);
     }
+
+    public function findRandomByUser(User $user)
```
doobry-systemli: PHPdoc annotations missing.

src/Repository/AliasRepository.php
```diff
@@ -39,4 +39,14 @@ public function findByUser(User $user)
...
+    {
+        return $this->findBy(['user' => $user, 'random' => true]);
+    }
+
+    public function findCustomByUser(User $user)
```
doobry-systemli: PHPdoc annotations missing.

src/Repository/AliasRepository.php
```diff
@@ -39,4 +39,14 @@ public function findByUser(User $user)
...
+    {
+        return $this->findBy(['user' => $user, 'random' => false]);
+    }
```
doobry-systemli: Do prevent code duplication, we could instead add an optional parameter `$random` to `findTypeByUser` and pass it to `findBy`?

```
    public function findTypeByUser(User $user, bool $random = false)
    {
        return $this->findBy(['user' => $user, 'random' => $random]);
    }
```

templates/Start/aliases.html.twig
```diff
@@ -1,14 +1,17 @@
...
     {{ form_errors(random_alias_form) }}
     <div class="form-group">
         {{ form_widget(random_alias_form.submit, {'attr': {'class': 'btn btn-primary' } }) }}
     </div>
     {{ form_end(random_alias_form) }}
 {% else %}
-    <p class="alert alert-info">{{ "index.alias-limit"|trans({'%alias_limit%': constant('App\\Handler\\AliasHandler::ALIAS_LIMIT')}) }}</p>
+    <p class="alert alert-info">{{ "index.alias-limit"|trans({
+            '%alias_limit%': constant('App\\Handler\\AliasHandler::ALIAS_LIMIT_RANDOM'),
+            '%alias_type': random|trans
```
doobry-systemli: missing `%` at the end of `%alias_type`?

src/Handler/AliasHandler.php
```diff
@@ -39,26 +40,44 @@ public function __construct(ObjectManager $manager, AliasCreator $creator)
...
      * @return Alias|null
      *
      * @throws ValidationException
      */
-    public function create(User $user, ?string $localPart): ?Alias
+    public function createRandom(User $user): ?Alias
     {
-        $aliases = $this->repository->findByUser($user);
-        if ($this->checkAliasLimit($aliases)) {
+        $aliases = $this->repository->findRandomByUser($user);
```
doobry-systemli: This check is not sufficient. It would allow a collision between an existing custom alias and a new random alias.

src/Handler/AliasHandler.php
```diff
@@ -39,26 +40,44 @@ public function __construct(ObjectManager $manager, AliasCreator $creator)
...
+     * @param string $localPart
+     *
+     * @return Alias|null
+     *
+     * @throws ValidationException
+     */
+    public function createCustom(User $user, string $localPart): ?Alias
+    {
+        $aliases = $this->repository->findCustomByUser($user);
+        if ($this->checkAliasLimit($aliases, self::ALIAS_LIMIT_CUSTOM)) {
```
doobry-systemli: This check is not sufficient. It would allow a collision between an existing random alias and a new custom alias.

src/Handler/AliasHandler.php
```diff
@@ -39,26 +40,44 @@ public function __construct(ObjectManager $manager, AliasCreator $creator)
...
-     * @param User        $user
-     * @param null|string $localPart
+     * @param User $user
      *
      * @return Alias|null
      *
      * @throws ValidationException
      */
-    public function create(User $user, ?string $localPart): ?Alias
+    public function createRandom(User $user): ?Alias
```
doobry-systemli: Why not keep a single method `create()` and use the condition `if (null === $localPart)` to distinguish between randomAlias and customAlias?

```
$limit = (null === $localPart) ? self::ALIAS_LIMIT_RANDOM : self::ALIAS_LIMIT_CUSTOM;
```

The `repository->find*ByUser()` condition needs to be unified anyway (see next two comments).

doobry-systemli:
> Right now, I expose the validator error message. On the good side, it shows why an alias couldn't be created. On the bad side:
> 
>     1. Allows to see if an email address already exists

> Number 1) is a bigger problem than in registration because users have the possibility to create and delete an infinite amount of aliases. Which makes enumeration easy. Do you have an idea how to handle this?

I don't see a big problem with enumeration. Also, it doesn't reveal whether the chosen name is in use or a reserved name. But it would

>     2. Error is always shown in English

This could be fixed: https://symfony.com/doc/current/validation/translations.html

doobry-systemli:
What do you think about changing the order? Put the custom alias stuff on top and the random alias stuff below? First, I think that custom aliases are more interesting for most users, second the random alias list can grow much bigger than the custom alias list.

t2d:
@doobry-systemli thank you for the thorough review. I tried to follow all your recommendations.

> What do you think about changing the order? Put the custom alias stuff on top and the random alias stuff below? First, I think that custom aliases are more interesting for most users, second the random alias list can grow much bigger than the custom alias list.

By now the order is as follow:
1) Button to add random alias
2) List of ALL aliases
3) Text field and button to add custom alias

Do you think we should split the list of aliases? I think it makes sense to keep it together since they are all just nicknames for your user in the end.

> I don't see a big problem with enumeration. Also, it doesn't reveal whether the chosen name is in use or a reserved name.

@0x46616c6b what do you think about the possibility of enumeration?

doobry-systemli:
> > What do you think about changing the order? Put the custom alias stuff on top and the random alias stuff below? First, I think that custom aliases are more interesting for most users, second the random alias list can grow much bigger than the custom alias list.
> 
> By now the order is as follow:
>     1. Button to add random alias
>     2. List of ALL aliases
>     3. Text field and button to add custom alias
> 
> Do you think we should split the list of aliases? I think it makes sense to keep it together since they are all just nicknames for your user in the end.

You're right, it totally makes sense to keep all aliases in a single list. I wonder whether we should move the form to add custom alias above the list to make it easier to access.

t2d:
> You're right, it totally makes sense to keep all aliases in a single list. I wonder whether we should move the form to add custom alias above the list to make it easier to access.

I tried this. But a) 2 buttons and 1 text field are confusing, and b) it doesn't look good

default_translations/de/messages.de.yml
```diff
@@ -10,7 +10,7 @@ index:
   voucher-headline: Deine Einladungscodes
   voucher-disable: Du bekommst am %date% Uhr drei Einladungscodes gutgeschrieben. Dies dient der Vermeidung von Massenregistrierungen und Spam.
   alias-headline: Deine Aliase
-  alias-limit: Du hast das Limit  von %alias_limit% erlaubten Aliases erreicht. Wenn du alte Aliases löschst, kannst du wieder neue erstellen.
+  alias-limit: Du hast das Limit  von %alias_limit% erlaubten %alias_type% Aliases erreicht. Wenn du alte Aliases löschst, kannst du wieder neue erstellen.
```
0x46616c6b: After `limit` is a double whitespace.

default_translations/de/messages.de.yml
```diff
@@ -35,12 +35,14 @@ form:
   voucher-redeemed-on: Eingelöst am
   oclock-by: Uhr von
   actual-password: Aktuelles Passwort
+  new-custom-alias: Neues spezifisches Alias
```
0x46616c6b: I would name it only `Neue Weiterleitung` or `Neues Alias`

src/Controller/StartController.php
```diff
@@ -155,11 +173,27 @@ private function createVoucher(Request $request, User $user)
...
+
+    /**
+     * @param Request $request
+     * @param User    $user
+     * @param string  $alias
+     */
+    private function createCustomAlias(Request $request, User $user, string $alias)
+    {
+        try {
+            if ($this->aliasHandler->create($user, strtolower($alias)) instanceof Alias) {
```
0x46616c6b: We should move the logic `strtolower` to the Creator to ensure that the usage of the creator is safe for any string you give as parameter.

src/Entity/Alias.php
```diff
@@ -26,12 +26,18 @@ class Alias implements SoftDeletableInterface
      */
     protected $destination;
 
+    /**
+     * @var bool
+     */
+    protected $random;
```
0x46616c6b: Extract the property `$random` to a Trait, make the property private and add a parameter for `setRandom`

default_translations/de/messages.de.yml
t2d: lets discuss this at a later point, when we have some feedback on the aliases

doobry-systemli: Why not change 'alias' to 'alias address' (en) and 'Alias' to 'Alias Adresse' (de) now? I think, that already helps a lot in understanding that it's about *mail addresses*.

t2d: Okay, i changed it to `alias address`

Issue #37, opened 2018-10-22
Change displayed message if user has only redeemed vouchers


doobry-systemli:
After we decided to only display active vouchers, users who have no active vouchers any longer now see the message that they will get new vouchers assigned. This needs to be fixed. We should distinguish between two cases:

* no vouchers are assigned yet (show the existing message)
* all vouchers have been used (show a new message)

PR #50, opened 2018-10-25
Display different message if all vouchers are used (Fixes: #37)

doobry-systemli:


default_translations/de/messages.de.yml
```diff
@@ -8,6 +8,7 @@ start:
 index:
   title: Verwalte deinen E-Mail Account
   voucher-headline: Deine Einladungscodes
+  voucher-limit: Du hast keine Vouchers mehr verfügbar.
```
0x46616c6b: We should use the term `Einladungscodes` like in the headline.

default_translations/en/messages.en.yml
```diff
@@ -8,6 +8,7 @@ start:
 index:
   title: Manage your email-account
   voucher-headline: Your invite codes
+  voucher-limit: You don't have any vouchers left.
```
0x46616c6b: We should use the term `invite code` like in the headline.

t2d:
@0x46616c6b good catch! I changed these and even more translations. Merge?

doobry-systemli:
Yep, well spotted. Should be merged.

doobry-systemli:
.. once the behat tests are fixed ;)

Issue #60, opened 2018-11-06
Alias should require either destination or user_id 


t2d:

t2d:
both need to be cleaned after soft deletion

Issue #61, opened 2018-11-06
Click on deleted alias in AliasAdmin produces NotFoundHttpException 


t2d:
# How to reproduce

1. Delete existing alias (f.e. 4)
2. Go to admin/alias/list (alias 4 is still listed)
3. Go to admin/alias/4/edit/

0x46616c6b:
I can't reproduce this anymore. Can you?

t2d:
I now get this page
![image](https://github.com/systemli/userli/assets/1834580/80282f9b-4976-4e22-a558-e50760c5296b)

t2d:
thank you. it's fixed now

PR #62, opened 2018-11-06
Alias improvements

### LINKED ISSUE #56: Aliases can be directly re-registred after deletion
If I delete an alias, another user can directly re-register it. I guess, that's not, what we wanted. Also, there are entries in the database left behind without `user_id`, which will probably never be deleted.

On the other hand:
If we allow users to create and delete aliases, but don't free up aliases after deletion,  1 single user could exhaust the complete namespace by just creating and deleting all (short) addresses.

t2d:
Fix #56 and #60

### DIFF
--- a/config/doctrine/Alias.orm.yml
+++ b/config/doctrine/Alias.orm.yml
@@ -20,7 +20,7 @@ App\Entity\Alias:
       nullable: false
     destination:
       type: string
-      nullable: false
+      nullable: true
     deleted:
       type: boolean
       options:
--- a/src/Admin/AliasAdmin.php
+++ b/src/Admin/AliasAdmin.php
@@ -13,6 +13,8 @@ use Symfony\Bridge\Doctrine\Form\Type\EntityType;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\EmailType;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+
 
 /**
  * @author louis <louis@systemli.org>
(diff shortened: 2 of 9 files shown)


src/Repository/AliasRepository.php
```diff
@@ -20,6 +20,16 @@ public function findOneBySource($email)
         return $this->findOneBy(array('source' => $email));
     }
 
+    /**
+     * @param $email
+     *
+     * @return array|Alias[]
+     */
+    public function findAllBySource($email)
```
0x46616c6b: We should add a second parameter for deleted which is default `false`

config/doctrine/Alias.orm.yml
```diff
@@ -20,7 +20,7 @@ App\Entity\Alias:
       nullable: false
     destination:
       type: string
-      nullable: false
+      nullable: true
```
0x46616c6b: We should not allow the destination to be empty, I guess postfix will use this field. We should ensure the field is filled by the user or we take the email from the user in there.

src/Admin/AliasAdmin.php
```diff
@@ -98,6 +119,10 @@ protected function configureBatchActions($actions)
     public function prePersist($alias)
     {
         if (null == $alias->getDestination()) {
+            if (null == $alias->getUser()) {
+                # set user_id to current user if neither destination nor user_id is given
+                $alias->setUser($this->storage->getToken()->getUser());
```
0x46616c6b: Here you should set the destination with the email from the user.

src/Entity/Alias.php
```diff
@@ -13,7 +13,7 @@
 /**
  * @author louis <louis@systemli.org>
  */
-class Alias implements SoftDeletableInterface
```
0x46616c6b: Why? This reverted the complete soft delete functionality.

config/doctrine/Alias.orm.yml
t2d: But it should be empty for privacy reasons after the alias has been deleted.

src/Entity/Alias.php
t2d: No, it doesn't. The actual functionality is in `DeleteTrait`.

src/Admin/AliasAdmin.php
t2d: That's excactly what I'm doing. But if no user is given, I'm getting the user first from the current admin. Afterwards the destination is still updated.

src/Entity/Alias.php
0x46616c6b: But we check in BaseRepository if the Class is instance of the Interface.

src/Admin/AliasAdmin.php
0x46616c6b: Ah okay, its a bit tricky to unterstand.

src/Entity/Alias.php
t2d: I see. I did this because of #61. Now it works similar to the User Entity.

t2d: lets do it in another PR

PR #85, opened 2018-12-19
WIP: Recovery process

doobry-systemli:


doobry-systemli:
Can you have a look and comment on whether you would implement things differently? If you like the general approach, I'll go ahead, add unit tests and implement the process to actually *use* the new recovery token.

t2d:
I like the process very much. It's exactly as I imagined it. Just some minor comments:

 *  You need page titles as well.
 * The UI could be improved:
    * When registering for the first time, I feel, the button should be below the pink line
button-position
    * When accessing `/user/recovery_token`:
      * I feel that parts of the strings `recovery-token.created-info` and  `recovery-token.displayed-once` should probably displayed permanently. Not only after creating a token. Otherwise, if I have already an account, I wouldn't know why I should create a recovery token.
      * The margin under the breadcrumbs has to be adjusted
top-margin

doobry-systemli:
will open a new one

Issue #83, opened 2018-12-17
Implement recovery token support


doobry-systemli:
We want recovery token support. The idea is as follows:

* At account creation, a recovery token is generated automatically. The user password is encrypted with this token and stored as `recoveryPasswordCipher` in database. The token is displayed to new users with strong emphasis to save it at a secure place.
* When users change their password, a new token will be generated automatically. Again, new token is displayed to the users.
* In account settings, users can change their recovery token. User password has to be entered in order to do so.

* When users lost their password, they can start the recovery process. They're asked to enter their recovery token. `recoveryPasswordCipher` is decrypted with the entered token and compared with the stored password. If both match, the entered token was correct. The recovery process is started.

Idea for the recovery process:

1. `recoveryProcessTime` is set to `NOW()`. In three days from now, but no later than one month, user can come back and reset their password. This process is explained to the user.
3. A mail is sent to the user mailbox that informs them about the process being started. Idea is to inform users that are still in charge of their account, when their token got stolen and the process got started by an attacker.
3. If user comes back (and enters correct recovery token) before three days have passed, the message that explains the process is displayed again.
4. If user comes back after three days, but no later than one month (and enters correct recovery token),  they're asked for a new password. Password changing process is started (set new password, generate new recovery token).
5. If user comes back later than one month (and enters correct recovery token), a fresh process is started.
6. (optional) a cronjobs clears `recoveryProcessTime` if set to a timestamp < NOW() - 1 month.

doobry-systemli:
The reason for the idea of a delayed recovery process is the following: the recovery token is enough to take over an account. Several scenarios are possible where a malicous person gets hold of the recovery token without the person in charge of the account noticing.

With the delayed process and notice mail, we provide an opportunity to the account holder to stop malicious recovery processes before the account is taken over.

PR #93, opened 2019-01-17
Implement recovery process (Partial #83)

doobry-systemli:
* Introduce recovery tokens in UUIDv4 (random) format
* Automatically generate a recovery token at account registration
* Allow to generate a new recovery token in account settings
* Store user password encrypted with recovery token in databae
* Implement process to change password using the recovery token:
  1. Need to authenticate with email + recovery token to start process
  2. Waiting period of 48 hours to mitigate hostile takeover
  3. Allow to set new password when re-authenticated with email +
     recovery token after 48 hours.

* Added new dependencies: 'ext-sodium', 'ramsey/uuid' and 'twig/extensions'
* Quick overview over the crypto process (using a combination of
  symmetric and asymmetric encryption):
  1. Derive a symmetric key from recovery token
  2. Deterministically generate a priv/pub keypair from derived key
  3. Encrypt user password with the public key
  4. Store pubkey along with the crypto cipher in database
  5. At password change, again encrypt new password with public key
  6. At recovery token change, go back to 1.
  7. To check recovery token, do 1. and 2. with provided token,
     decrypt cipher with the generated key and compare the decrypted
     string with the stored password.

* TODO:
  * Track recovery token state. Three possible states:
    1. Legacy users (no token generated)
    2. Pending (token generated but not acknowledged or token being used
       in process)
    3. Token set and acknowledged
    Idea is to force users in state 'pending' to generate and acknowledge
    a new token, overriding any routes to other pages.
  * Optionally auto-complete default domain for username at token verification?
  * Send out mail to user about recovery process being started.
  * Implement deep link to reset the recovery process, link it in the
    notification mail.


src/Form/RegistrationRecoveryTokenType.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+class RegistrationRecoveryTokenType extends AbstractType
+{
+    const NAME = 'registration_recovery_token';
+
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('ack', CheckboxType::class, [
+                'required' => false,
```
0x46616c6b: This must be `true` otherwise the Checkbox is optional and validation is only made with ConstraintValidator.

templates/Registration/recovery_token.html.twig
```diff
@@ -0,0 +1,41 @@
...
+{% block content %}
+    <div class="row">
+        <div class="col-xs-12 col-sm-6 col-md-6">
+            <h3>{{ "recovery-token.headline"|trans }}</h3>
+        </div>
+    </div>
+    <div class="row">
+        <div class="col-xs-12 col-sm-6 col-md-6">
+            <p class="lead">{{ "recovery-token.created-headline"|trans }}</p>
+            <p class="lead alert alert-warning">{{ recovery_token }}</p>
```
0x46616c6b: This should be only a warning (yellow) when the user don't have a recovery_token, otherwise it should be green.

default_translations/de/messages.de.yml
```diff
@@ -53,6 +55,19 @@ form:
   create-custom-alias: Erstelle Alias-Adresse
   create-random-alias: Generiere zufällige Alias-Adresse
   delete-alias: Alias-Adresse löschen
+  generate-recovery-token: Neuen Wiederherstellungscode generieren
+  recovery-token: Wiederherstellungscode
+  recovery-start: Prozess starten
```
0x46616c6b: I would only call it "Zurücksetzen/Reset". "Process start" sound very technical and even not what the user want.

config/validator/validation.yaml
```diff
@@ -40,6 +40,12 @@ App\Form\Model\Registration:
         plainPassword:
             - App\Validator\Constraints\PasswordPolicy: ~
 
+App\Form\Model\RecoveryTokenAck:
```
0x46616c6b: I would call this Form and Model `RegistrationRecoveryToken`

src/Controller/RecoveryController.php
```diff
@@ -0,0 +1,300 @@
...
+    /**
+     * @param Request $request
+     *
+     * @return Response
+     *
+     * @throws \Exception
+     */
+    public function recoveryProcessAction(Request $request): Response
+    {
+        $processState = 'NONE';
```
0x46616c6b: This should be a constant.

src/Controller/RecoveryController.php
```diff
@@ -0,0 +1,300 @@
...
+            ]
+        );
+    }
+
+    /**
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function recoveryTokenAckAction(Request $request): Response
```
0x46616c6b: This is related to the registration process, right? Then we should move this action to the RegistrationController.

src/Form/Model/RecoveryResetPassword.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    /**
+     * @var string
+     */
+    public $email;
+
+    /**
+     * @var string
+     */
+    public $recoveryToken;
```
0x46616c6b: When you use getter/setter (i prefer that) then you should use only private properties.

src/Form/RegistrationRecoveryTokenType.php
doobry-systemli: Actually, that's what I wanted to achieve. I didn't like the HTML5 error layout, so I did it that way to instead display the validation error from Symfony ConstraintValidator.

config/validator/validation.yaml
doobry-systemli: It's used both by RegistrationController and by RecoveryController, so I decided to give it a more general name.

src/Controller/RecoveryController.php
doobry-systemli: but `$processState` changes depending on the process state. Or do you mean constants for all states and using the relevant one then?

0x46616c6b: Of course for all the States.

config/validator/validation.yaml
0x46616c6b: But you named it now for the only property not for the purpose of the form.

src/Controller/RecoveryController.php
doobry-systemli: Nope, it's the form that checks whether the token has been acknowledged. This happens *both* in registration process and in user settings -> recovery token. That's why there's `registerRecoveryTokenAckAction` in RegistrationController and `recoveryTokenAckAction` in RecoveryController.

templates/Registration/recovery_token.html.twig
doobry-systemli: Agreed, fixed now.

default_translations/de/messages.de.yml
doobry-systemli: Agreed, fixed now.

src/Form/Model/RecoveryResetPassword.php
doobry-systemli: Agreed, fixed now.

src/Controller/RecoveryController.php
doobry-systemli: I see. I changed them to constants now.

config/validator/validation.yaml
doobry-systemli: Mh, not sure whether I understand your point. The model is called `RecoveryTokenAck` because it's a model for the form to acknowledge (ack) the recovery token. What's wrong with the name in your opinion?

src/Form/RegistrationRecoveryTokenType.php
doobry-systemli: Done now.

Issue #117, opened 2019-02-05
Allow to pass invite code as parameter


t2d:
It would be great if I could send my friend a link like https://users.example.org/register&v=sadfas and this would already fill out the "invite code" in the register field.
We would also need an option to create the link in a sensible way which doesn't create confusion.

graybee:
Is threre anyone there who can give me an invite code for SYSTEMLI.ORG email. I would be very thankful coz your so kind and generous. My email is Tinbox404-hot@yahoo.com

PR #132, opened 2019-02-19
Add support for prefilled registration links (Fixes: #117)

doobry-systemli:
* Add new route for registration links with prefilled invite vouchers.
* Make vouchers in the voucher list to hyperlinks to those registration
  links.
* Add new commandline options for printing vouchers and voucher links to
  the VoucherCreation command.
* Add behat tests for voucherCreationCommand.

### DIFF
--- a/config/routes.yaml
+++ b/config/routes.yaml
@@ -52,6 +52,10 @@ register_welcome:
     path:     /register/welcome
     controller: App\Controller\RegistrationController::welcomeAction
 
+register_voucher:
+    path:     /register/voucher/{voucher}
+    controller: App\Controller\RegistrationController::registerAction
+
 # RecoveryController
 
 recovery:
--- a/config/routes.yaml
+++ b/config/routes.yaml
@@ -53,7 +53,7 @@ register_welcome:
     controller: App\Controller\RegistrationController::welcomeAction
 
 register_voucher:
-    path:     /register/voucher/{voucher}
+    path:     /register/{voucher}
     controller: App\Controller\RegistrationController::registerAction
 
 # RecoveryController
--- a/src/Form/RegistrationType.php
+++ b/src/Form/RegistrationType.php
@@ -39,7 +39,6 @@ class RegistrationType extends AbstractType
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        dump($options['data']->getVoucher());
         $transformer = new TextToEmailTransformer($this->domain);
 
         $builder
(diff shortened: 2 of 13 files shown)


0x46616c6b:
Why we need a new route for tha? We can transport the parameters as query strings and check that in the existing controller and extend the template for having prefilled form.

templates/Start/vouchers.html.twig
```diff
@@ -31,7 +31,7 @@
                             </div>
                         {% else %}
                             <div class="alert alert-success">
-                                <p class="voucher-box">{{ voucher.code }}</p>
+                                <p class="voucher-box"><a href="{{ url('register_voucher', {'voucher': voucher.code}) }}" target="_blank">{{ voucher.code }}</a></p>
```
0x46616c6b: We should add a "click to copy link" button and not use a link.

src/Form/RegistrationType.php
```diff
@@ -39,10 +39,14 @@ public function __construct(string $domain)
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
+        dump($options['data']->getVoucher());
```
0x46616c6b: Need to be removed.

templates/Start/vouchers.html.twig
doobry-systemli: Agreed, but I don't have clue how to do that. Would you be ok with doing this later in a separate step?

src/Form/RegistrationType.php
doobry-systemli: Thanks, done.

config/routes.yaml
```diff
@@ -52,6 +52,10 @@ register_welcome:
     path:     /register/welcome
     controller: App\Controller\RegistrationController::welcomeAction
 
+register_voucher:
+    path:     /register/voucher/{voucher}
```
0x46616c6b: Why not only `/register/{voucher}`?

doobry-systemli: I would prefer `/register/{voucher}` myself. I chose `register/voucher/{voucher}` to prevent possible colisions with `/register/welcome` and `/register/recovery_token`. Collisions are impossible at the moment as vouchers always have a lenght of six characters, but that might change in the future. I woul be fine with changing the route back to `/register/{voucher}` though.

0x46616c6b: I would prefer that route. So it is also the last route with `/register/*`, even if their is a voucher called welcome they never can use the invite link not the welcome page is not working.

doobry-systemli: Ok, changed route to `/register/{voucher}`

Issue #108, opened 2019-02-01
[aliases] Send mail to user at alias creation


doobry-systemli:
We should send a mail to the user account when they create a new alias.

PR #136, opened 2019-02-21
Send mail to user at alias creation (Fixes: #108)

doobry-systemli:


default_translations/de/messages.de.yml
```diff
@@ -212,3 +201,37 @@ Bad credentials.: Fehlerhafte Zugangsdaten
 
 custom: gewählten
 random: zufälligen
+
+mail:
```
t2d: When you touch translations anyway, please use the opportunity to break lines at a reasonable length.

Issue #130, opened 2019-02-15
Missing description for password policy on Account Page


0x46616c6b:
If I want to change my password and not fulfil the password policy I expect that I get a hint what the policy looks like.

![screenshot_2019-02-15 example org - account verwalten](https://user-images.githubusercontent.com/363667/52854583-22da7480-311f-11e9-9e64-56303a4f6edd.png)

PR #137, opened 2019-02-21
Add infobox with password policy to password change forms (Fixes: #130)

doobry-systemli:


doobry-systemli:
screenshot_2019-02-22 example org - recovery reset-header

screenshot_2019-02-22 example org - account verwalten

PR #138, opened 2019-02-21
Adjust database to snake_case (Fix #112)

### LINKED ISSUE #112: Normalize database fields
I believe we want all database fields in snake_case?! We should do this with 2.0.0 since this  probably breaks compatibility.

t2d:


t2d:
@doobry-systemli  I think Github Markdown doesn't recognize the markup correctly you used before in `UPGRADE.md`.

doobry-systemli:
> I think Github Markdown doesn't recognize the markup correctly you used before in UPGRADE.md.

The indented format for code blocks is the same we use in `README.md`. For plaintext view (e.g. in editors) I prefer it over the format with 3x `. Anyway, we should harmonize the style in both files, I don't have strong feelings about it.

Apart from that, the PR looks good to me!

Issue #106, opened 2019-01-29
[recovery] Invalidate token when used and generate new one


doobry-systemli:
When a recovery process was successful, the used token should be deleted. When user logs in, they should be presented a new token that they have to acknowledge as usual.

PR #140, opened 2019-02-22
[WIP] Renew recovery token after usage (Fixes: #106)

doobry-systemli:
* Invalidate old recovery token when it has been used
* Create new recovery token at the end of recovery process

### DIFF
--- a/config/routes.yaml
+++ b/config/routes.yaml
@@ -66,6 +66,10 @@ recovery_reset_password:
     path:     /recovery/reset_password
     controller: App\Controller\RecoveryController::recoveryResetPasswordAction
 
+recovery_recovery_token_ack:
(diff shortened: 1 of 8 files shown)


doobry-systemli:
>  You can just abort the process after resetting the password (at http://192.168.33.99/recovery/reset_password). Then the old code seems to be still valid.

Nope, that should not be true. The token is invalidated and a new one generated in the same method as the password reset. See here: https://github.com/systemli/userli/blob/recovery_token_refresh/src/Controller/RecoveryController.php#L400-L413

t2d:
Tab 1: screenshot from 2019-02-23 10-29-39

Tab 2: 
screenshot from 2019-02-23 10-29-51



Well, than it's even worse. We tell the user that they have a recovery token, but they actually don't.

PR #144, opened 2019-03-04
add filter and column for recovery token usage

0x46616c6b:
As Admin I want to see if a user has set a recovery token or not. Also I want to filter the users for this criteria.

### DIFF
--- a/src/Admin/UserAdmin.php
+++ b/src/Admin/UserAdmin.php
@@ -12,6 +12,7 @@ use Sonata\AdminBundle\Datagrid\DatagridMapper;
 use Sonata\AdminBundle\Datagrid\ListMapper;
 use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
+use Sonata\Form\Type\BooleanType;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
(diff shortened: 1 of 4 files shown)


src/Admin/UserAdmin.php
```diff
@@ -152,6 +167,9 @@ protected function configureListFields(ListMapper $listMapper)
             ->addIdentifier('email')
             ->add('creationTime')
             ->add('updatedTime')
+            ->add('hasRecoverySecretBox', 'boolean', [
+                'label' => 'Recovery Token',
```
doobry-systemli: I can imagine that this label causes the column to become very wide and take a lot of space? That would be unfortunate for a boolean column which has only 0/1 as values. Please ignore the comment if my assumption is wrong.

0x46616c6b:
screenshot_2019-03-05 admin - user list

src/Admin/UserAdmin.php
0x46616c6b: I added a screenshot in the comments here. So you can get a feeling for the visual change.

t2d:
Why can I sort by every other column but not "Recovery Token"?

0x46616c6b:
screenshot_2019-03-05 admin - user list 2

t2d:
I added some code to export the same information to Munin. Hope it's okay to do it with this PR. Very similar idea.

PR #149, opened 2019-03-06
Accept recovery email also with local part only (fix #148)

### LINKED ISSUE #148: Accept local part of email address in /recovery
We accept it everywhere else, if it's from primary domain

t2d:


src/Form/DataTransformer/OptionalDomainEmailTransformer.php
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function transform($value)
+    {
+        if (null === $value) {
+            return '';
+        }
+
+        if (false !== $at_position = stripos($value, '@')) {
```
0x46616c6b: CamelCase please

tests/Form/DataTransformer/OptionalDomainEmailTransformerTest.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace App\Tests\Form\DataTransformer;
+
+use App\Form\DataTransformer\OptionalDomainEmailTransformer;
+use PHPUnit\Framework\TestCase;
+
+class OptionalDomainEmailTransformerTest extends TestCase
+{
+    const DOMAIN = 'systemli.org';
```
doobry-systemli: Should be 'example.org' for consistency reasons.

tests/Form/DataTransformer/OptionalDomainEmailTransformerTest.php
```diff
@@ -0,0 +1,55 @@
...
+        $this->assertEquals($expected, $this->getTransformer()->reverseTransform($input));
+    }
+
+    public function reverseTransformProvider()
+    {
+        return [
+            ['', ''],
+            [null, ''],
+            ['louis', 'louis@systemli.org'],
+            ['louis@example.org', 'louis@example.org'],
```
doobry-systemli: Not sure whether we should use real adresses in the tests. I know that it has been like that in the past, but ...

config/services.yaml
```diff
@@ -74,6 +74,12 @@ services:
         tags:
             - { name: form.type, alias: password_change }
 
+    App\Form\RecoveryProcessType:
+        arguments:
+            $domain: '%env(DOMAIN)%'
+        tags:
+            - { name: form.type, alias: recovery }
```
doobry-systemli: I don't think that the tag is necessary here, is it?

PR #158, opened 2019-03-13
New dotenv handlin

t2d:
Merge after #157 

I did the steps explained in https://symfony.com/doc/current/configuration/dot-env-changes.html
(more or less)


doobry-systemli:
From this PR I don't understand how the new dotenv handling will look like. Do we just have to add a `.env.local` file with our custom settings in deployment process and that's it? Would be nice to have this documented in any case.

PR #170, opened 2019-05-11
Add pt_BR translation files

bruno-gama:
Addressing issue https://github.com/systemli/userli/issues/168

This adds the translation files for Brazilian Portuguese. I've made an effort to keep the translation as gender neutral as possible. 

Notes:

I've placed the translation files in a new `default_translations/pt_br` folder, but I'm actually not sure if thats the right folder of if it should be called `pt_br` or just `pt`, I figure it's better to just open the PR and then ask than trying to figure it out on my own.

### DIFF
--- a/default_translations/pt/messages.pt.yml
+++ b/default_translations/pt/messages.pt.yml
@@ -1,0 +1,257 @@
+start:
+  title: Te damos as boas-vindas a %domain%!
+  intro: Aqui você pode criar e configurar a sua conta de e-mail.
+  registration-header: Registro
+  registration-text: >
+    Ainda não tem uma conta em %domain%? Sem problemas!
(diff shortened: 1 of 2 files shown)


t2d:
Hey Bruno,

Thanks a lot. This looks fantastic. And thank you for keeping it gender neutral.
As this is the only Portuguese translation it would be great to move it to `default_translations/pt/`. Can you do it or do you want me to do it? Do you want to start a contributors file and add your name to it? Or do you don't mind?

bruno-gama:
No problem I can rename the folder, will do that later today.

Also I'm indifferent about a contributors file at least for now. I hope I can find the time to push some code in the future, and maybe then we can see about that.

PR #174, opened 2019-06-18
Rework checkpassword

doobry-systemli:
* Accept partial input stream (don't expect exactly four fields separated by ('\0')
 * Get mailCrypt fields if mailCrypt is enabled for the user in question
 * Better error handling if $email or $password are empty
 * Update documentation accordingly


t2d:
I have tested this branch and integrated it into `ansible-role-dovecot`. It seems to work nicely. I have not looked at the concerning class so far.

Issue #165, opened 2019-04-18
Copy button for invitation links is broken


0x46616c6b:
**Steps to reproduce**

- Log in and go to `/voucher`
- Click on gray button inside the voucher item
- Clipboard contains `undefined`

**Expected**

After clicking on the button I expect to get the invite url.

t2d:
Can't reproduce. It works for me.

0x46616c6b:
With deactivated uBlock Origin (Firefox 66.0.2) it works also for me.

t2d:
For me it works with Firefox and ublock origin. Maybe a specific rule set? Can this be closed?

doobry-systemli:
For me this definitely was broken as well a few days ago (with ublock origin). Now it's working again (with ublock origin). No idea what changed.

@0x46616c6b: is it still broken for you?

t2d:
Haha, I was able to find the bug. The link is undefined if you click on the glyphicon, but it works if you click on the button surrounding it. 

```
el.value = element.target.dataset.link;
```
Doesn't seem to get the right target if clicking directly on the icon.

PR #194, opened 2019-11-09
WIP: add a button to copy random aliases to clipbard

### LINKED ISSUE #181: Add copy button to random aliases also
Copy the invite code copy button to random aliases.

jelhan:
Also refactors the existing logic to copy content to clipboard for reusability. A copy button could be created with `<button data-button="copy-to-clipboard" data-value="foo">`. This will copy `"foo"` to clipboard when user clicks the button. `[data-button="copy-to-clipboard"]` is used as selector to attach the event listener.

This also addresses #165 in another way. b5ba3f9a341f1fbab5cbb19ea5fafcad65770d32 required to duplicate the content that should be copied on each child element. This could be avoided by using `Event.currentTarget` instead of `Event.target`. `Event.currentTarget` is supported by IE9 and above.

Closes #181

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -3,16 +3,24 @@ require('../../vendor/twbs/bootstrap/dist/js/bootstrap.min.js');
 require('../css/app.css');
 
 $(function () {
-    $('[data-toggle="tooltip"]').tooltip();
-    $('[data-toggle="popover"]').popover();
-
-    // We create a non visible textarea with the link,
-    // select them and tell the browser to copy the selected content
-    // after we copied the link the element will removed from DOM.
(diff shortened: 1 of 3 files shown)


jelhan:
This adds a new translation key: `index.alias-copy` I wasn't sure what is required to do so. Should I add it manually to Weblate? Or to the translations stored in `default_translations` folder? This is the reason that this one is still WIP.

t2d:
Thank you very much for your PR. I merged it manually.
Concerning the strings, I replaced both with a generic "Copy to clipboard" string. I think, this is the most effective solution atm. To add new strings just add them to our main language files `default_translations/en/messages.en.yaml`

PR #196, opened 2019-11-25
Add route to set admin password

### LINKED ISSUE #195: Create first domain and account
Currently after installation, there is a database scheme, but no data in it. So it's impossible to login. We should either fill in some standard data or create a mechanism to create a first domain and account.

Came up in #193

t2d:
* Redirect from index if no domain is configured
* Redirect to index if domain is configured
* Fix #195


t2d:
Can you please give me feedback if you like the idea? Or do you think a CLI would be the better choice here? Then I'll create more translations and tests.

0x46616c6b:
I like the idea!

t2d:
I have implemented a new route `init` where the first visitor is redirected to, to create the primary domain and set the password of `admin@primary domain`.

Furthermore, there is the command `app:admin:password`, that is executable all the time to (re-)set the admin password. That implies, there everyone who can access the command line, can reset the password. That's probably a security problem, isn't it?
Before only users with mysql access could reset the password.
Should I copy the behavior, that `app:admin;password` is only accessible when no account exists?

Both interfaces use the same helper function and are covered by tests.

All feedback very welcome!

doobry-systemli:
I reviewed the PR and I like it as well. Thanks a lot for your work!

> Furthermore, there is the command `app:admin:password`, that is executable all the time to (re-)set the admin password. That implies, there everyone who can access the command line, can reset the password. That's probably a security problem, isn't it?
> Before only users with mysql access could reset the password.
> Should I copy the behavior, that `app:admin;password` is only accessible when no account exists?

I don't think this is a security problem. Users with console access can read `.env` with the database password anyway, no? And I definitely like the idea of being able to reset the admin password at a later point. But we would have to limit that function to admin accounts without mail_crypt since otherwise we invalidate the mail_crypt key pair by setting a new password. In other words, we would have to check for condition `!$user->hasMailCryptSecretBox()` before changing the password.

PR #252, opened 2020-10-09
Web Key Directory support

doobry-systemli:
This PR adds support for lettings users upload their OpenPGP key and exporting them to a Web Key Directory in the format as described in https://tools.ietf.org/html/draft-koch-openpgp-webkey-service-10.

overview

openpgp-settings


doobry-systemli:
And tests are also missing :wink:

doobry-systemli:
Apart from tests, this PR is ready to be merged from my side.

doobry-systemli:
Behat tests for the feature are also added now.

I don't consider it worth the time to write unit tests for all the functions since for now the functions are only used by this particular feature anyway. Therefore functional tests are sufficient in my eyes.

doobry-systemli:
Mh, can somebody shed some light on why the `codecov/patch` and `codecov/project` CI jobs are marked as failed? Sure, code coverage degreased, but only when looking at PHPunit tests. Looking at behat functional tests, code coverage even *increased*.

I couldn't find any hints in the failed jobs about how to make them pass :cry:

0x46616c6b:
> Mh, can somebody shed some light on why the `codecov/patch` and `codecov/project` CI jobs are marked as failed? Sure, code coverage degreased, but only when looking at PHPunit tests. Looking at behat functional tests, code coverage even _increased_.
> 
> I couldn't find any hints in the failed jobs about how to make them pass 😢

Failed jobs are not a blocker to merge the PR. If the coverage decreases the job will fail. In my eyes this is okay and I'm a fan of not stopping the progess to have a higher coverage. Decreasing of coverage will rise the risk to not catch regressions. Also know that writing Unittests is not so lovely like feature or E2E tests but would love to have some here.

t2d:
> Mh, can somebody shed some light on why the `codecov/patch` and `codecov/project` CI jobs are marked as failed? Sure, code coverage degreased, but only when looking at PHPunit tests. Looking at behat functional tests, code coverage even _increased_.
> 
> I couldn't find any hints in the failed jobs about how to make them pass cry

codecov only looks at phpunit. let's just ignore it in this case. I don't agree that tests are not necessary, but I see that writing them is tedious.

UPGRADE.md
```diff
@@ -1,5 +1,12 @@
 # Upgrade documentation
 
+## Upgrade from 2.6.1 or lower
+
+Database schema changed and needs to be updated:
+
+    ALTER TABLE virtual_users
+    ADD wkd_key LONGTEXT DEFAULT NULL;
```
t2d, line 8: You changed the format of the key to binary. Would it make sense to also adjust the database field type?

doobry-systemli: I decided to go with base64 encoded binary instead to not have binary blobs in the database. Would you prefer binary blobs in database? If so, why?

config/doctrine/User.orm.yml
```diff
@@ -51,6 +51,9 @@ App\Entity\User:
     mailCryptPublicKey:
       type: text
       nullable: true
+    wkdKey:
+      type: text
+      nullable: true
```
t2d, line 56: See the comment above about binary field types, especially `blob`
https://www.doctrine-project.org/projects/doctrine-dbal/en/2.11/reference/types.html#blob

doobry-systemli:
> codecov only looks at phpunit. let's just ignore it in this case. I don't agree that tests are not necessary, but I see that writing them is tedious.

I didn't mean that unit tests aren't useful. My point rather was that it's the wrong approach to aim for code coverage numbers. For some functions, unit tests simply don't make sense. If the functions do have actual business code to test for - and especially if the functions are used in more than one places in code, I absolutely agree that unit tests are helpful!

I just added unit tests for `OpenPGPWkdHandler` class (and for `WKDDeleteType` as that was very simple).

t2d:
Padlock image doesn't load in my vagrant
image

```
GEThttp://192.168.33.99/build/images/padlock.svg
[HTTP/1.1 404 Not Found 477ms]

	
GET
	http://192.168.33.99/build/images/padlock.svg
Status404
Not Found
VersionHTTP/1.1
Transferred42.88 KB (273.54 KB size)

    	
    Cache-Control
    	no-cache, private
    Connection
    	keep-alive
    Content-Encoding
    	gzip
    Content-Type
    	text/html; charset=UTF-8
    Date
    	Thu, 15 Oct 2020 14:04:27 GMT
    Referrer-Policy
    	no-referrer, strict-origin-when-cross-origin
    Server
    	nginx/1.14.2
    Transfer-Encoding
    	chunked
    X-Content-Security-Policy
    	default-src 'none' 'self'; block-all-mixed-content ; font-src 'self'; img-src 'self'; script-src 'self' 'unsafe-inline' 'nonce-fd35d22053a54cf3cda960f0c5944cc9'; style-src 'self' 'unsafe-inline' 'nonce-df18aebbe88ea6cb98648a83cc407659'
    X-Content-Type-Options
    	nosniff
    X-Debug-Token
    	8b6d44
    X-Debug-Token-Link
    	http://192.168.33.99/_profiler/8b6d44
    X-Frame-Options
    	DENY
    X-Previous-Debug-Token
    	41b1a7
    X-Robots-Tag
    	noindex
    X-XSS-Protection
    	1; mode=block
    	
    Accept
    	image/webp,*/*
    Accept-Encoding
    	gzip, deflate
    Accept-Language
    	en
    Connection
    	keep-alive
    Cookie
    	PHPSESSID=ssdeu5clnf32i5vfagpihe9avm
    DNT
    	1
    Host
    	192.168.33.99
    Referer
    	http://192.168.33.99/fe/
    Sec-GPC
    	1
```

doobry-systemli:
> Padlock image doesn't load in my vagrant

I guess that you need to run some build script in order to get that copied, but to be honest I don't remember which one :grimacing:

t2d (commented):
> > Padlock image doesn't load in my vagrant
> 
> I guess that you need to run some build script in order to get that copied, but to be honest I don't remember which one grimacing

`yarn encore dev`

t2d:
It would be great if you could adjust the ansible playbook so that the necessary directory is created. I failed to do it manually and couldn't test further.

doobry-systemli:
> It would be great if you could adjust the ansible playbook so that the necessary directory is created. I failed to do it manually and couldn't test further.

Mh, if permissions allow it, `WKD_DIRECTORY` should be created automatically when importing a key. And as the webserver is run by `www-data` and the default path for `WKD_DIRECTORY` is under `/var/www/html` (which usually is owned by `www-data` as well), this should work - at least it did in my tests :thinking:

Anyway the following should be sufficient for further testing:

```
mkdir -p /var/www/html/.well-known/openpgpkey
chown www-data:www-data github.com/systemli/ansible-role-userli
```

You can use the provided console commands to test WKD functionality, just take care to run them as `www-data` in order to allow further testing via webinterface:

```
gpg --gen-key # generate key, e.g. for admin@example.org
gpg --export <KEY_ID> >/tmp/key.asc

sudo -u www-data ./bin/console app:users:import-key admin@example.org /tmp/key.asc
sudo -u www-data  ./bin/console app:users:show-key admin@example.org
find /var/www/html/.well-known/openpgpkey

sudo -u www-data ./bin/console app:users:delete-key admin@example.org
find /var/www/html/.well-known/openpgpkey
```

and for reference also:

```
sudo -u www-data ./bin/console app:users:wkd:export-keys
```

default_translations/de/messages.de.yml
```diff
@@ -17,6 +17,12 @@ start:
   vouchers-desc: Lade Freund*innen ein
   webmail: Webmail
   webmail-desc: Lies und schreib E-Mails
+  wkd: OpenPGP Schlüssel
+  wkd-delete: OpenPGP Schlüssel löschen
+  wkd-settings: OpenPGP Schlüssel
```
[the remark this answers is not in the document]
doobry-systemli: I agree and now merged them.

doobry-systemli:
So I went through all the points identified by @0x46616c6b in our peer review and did a lot of code refactoring. Major changes:

* Route is changed from `/wkd` to `/openpgp` to reflect the fact that we're actually handling OpenPGP keys and exporting them to WKD is only one feature - more may follow :wink: 
* Key expiry date is stored in the database and displayed in webinterface.
* Keys are removed both from database *and* from WKD when user is deleted.
* `OpenPgpKey` is an own entity with an own database table this also allows to add OpenPGP keys that are not tied to a user (e.g. OpenPGP keys of Schleuder lists).
* `WkdHandler` uses a `GpgKeyImporter` with static functions that uses a `OpenPgpKeyImporterInterface`.
* `GpgKeyImporter` filters the UIDs of imported keys and throws away all UIDs that don't contain the email address in question.

doobry-systemli:
In my eyes, this PR is ready to be merged. Any objections? @0x46616c6b and @t2d would you take another look?

PR #270, opened 2020-11-19
Change input type to make css work

1resu:
The css file uses `.form-signin input[type="email"]` - [app.css](https://github.com/systemli/userli/blob/57cbe34a7bcb3f00324977a4f45b299c3562b631/assets/css/app.css#L35)

### DIFF
--- a/templates/Security/login.html.twig
+++ b/templates/Security/login.html.twig
@@ -11,7 +11,7 @@
                 {% if error %}
                     <div class="alert alert-danger" role="alert">{{ error.message|trans }}</div>
                 {% endif %}
-                <input type="text" id="username" class="form-control" name="_username" placeholder="{{ "form.email"|trans }}" value="{{ last_username }}" required="" autofocus="" />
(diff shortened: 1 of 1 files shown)


doobry-systemli:
Can you elaborate on what you try to achieve here? Seems legit to change the input type to `email`, but I'd like to understand what's going to be fixed with this change.

Also, if you change the input field type in the login form on login page, it should be changed in the login form on startpage accordingly.

1resu:
> Can you elaborate on what you try to achieve here?

Please have a look at [this](https://users.systemli.org/de/login) page. The space between doesn't look right. And also the non-rounded corners of the password field. There is not css style for an input field of the type "text".

I assumed you mistakenly choosed the type of the input field. With this commit the form fields at /login look much nicer.

doobry-systemli:
I see, thanks a lot for the explanation. In my browsers, the visual changes are marginal, but I agree that 'email' is the correct input field type here :blush:

t2d:
This result in users now being required to enter full email addresses, not only localpart. I think it's not super critical, but I think it wasn't intended

PR #336, opened 2021-12-12
Timestamps for entities

### LINKED ISSUE #207: Set creationTime and updatedTime in all entity constructors

glitchtim:
This should close #207.

**Notes:**

The entity OpenPgpKey wasn't changed, as there are some tests using direct object comparison, which fail in *App\Tests\Handler\WkdHandlerTest*, if the creation date is set in the constructor (as they differ by just a few microseconds).

This might be fixed by setting the fields creationTime and updatedTime in the affected tests in *WkdHandlerTest*.

### DIFF
--- a/src/Entity/Alias.php
+++ b/src/Entity/Alias.php
@@ -37,7 +37,9 @@ class Alias implements SoftDeletableInterface
     {
         $this->deleted = false;
         $this->random = false;
-        $this->creationTime = new \DateTime();
+        $currentDateTime = new \DateTime();
+        $this->creationTime = $currentDateTime;
+        $this->updatedTime = $currentDateTime;
     }
 
     /**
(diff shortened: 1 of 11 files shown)


doobry-systemli:
Thanks a lot @glitchtim for your contribution <3

Given that `creationDate` and `updatedDate` are set in the entity constructors now, we should remove the code lines where they're set again explicitely after a new entity got instantiated. I found https://github.com/systemli/userli/blob/3840d4afcb4f0842acd27331dfdee80f463ebedb/src/Factory/DomainFactory.php#L14-L15 and https://github.com/systemli/userli/blob/3840d4afcb4f0842acd27331dfdee80f463ebedb/src/Handler/RegistrationHandler.php#L131-L132.

> The entity OpenPgpKey wasn't changed, as there are some tests using direct object comparison, which fail in App\Tests\Handler\WkdHandlerTest, if the creation date is set in the constructor (as they differ by just a few microseconds).
> 
> This might be fixed by setting the fields creationTime and updatedTime in the affected tests in WkdHandlerTest.

We should set the dates in the OpenPgpKey constructor as well. I would be fine with explicitely re-setting the times in our tests if we do direct object comparison.

PR #340, opened 2021-12-31
Add cron job and link to maildirs-remove script

1resu:
This addition explains how to delete maildirs from removed users. Taken from the Userli Ansible role.

### DIFF
--- a/hugo/content/installation/finalize.md
+++ b/hugo/content/installation/finalize.md
@@ -27,10 +27,13 @@ the `mail_location` in `10-mail.conf` to something like this:
 
 ## Cronjobs
 
-Some cronjobs are needed in order to run regular tasks:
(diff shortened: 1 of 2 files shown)


hugo/content/installation/finalize.md
```diff
@@ -27,10 +27,13 @@ the `mail_location` in `10-mail.conf` to something like this:
 
 ## Cronjobs
 
-Some cronjobs are needed in order to run regular tasks:
+Some cronjobs are needed in order to run regular tasks. As Userli does not have write permissions at Dovecot's maildir (usually this directory belongs to the system user `vmail`) you have to use a [script](https://github.com/systemli/ansible-role-userli/blob/master/templates/userli-maildirs-remove.sh.j2) to delete a  maildir from a removed Userli account:
```
doobry-systemli, line 30: So far, we tried to keep lines below 80 chars in the docs.

```suggestion
Some cronjobs are needed in order to run regular tasks. As Userli does not
have write permissions at Dovecot's maildir (usually this directory belongs
to the system user `vmail`) you have to use a [script](https://github.com/systemli/ansible-role-userli/blob/master/templates/userli-maildirs-remove.sh.j2)
to delete a maildir from a removed Userli account:
```

doobry-systemli (changes requested):
Thanks a lot for your contribution! If you accept the line breaks, it's good to be merged :)

PR #364, opened 2022-04-08
:bug: Fix nullable $domain in WelcomeMessage*.php

0x46616c6b:
I received this errors when I started a empty userli instance and wants to bootstrap the system.

```
In WelcomeMessageSender.php line 35:
                                               
  Call to a member function getName() on null  
                                               
```

```
In WelcomeMessageBuilder.php line 36:
                                               
  Call to a member function getName() on null  
                                               
```


t2d:
Apart from the comment this looks great. Thanks a lot!

PR #367, opened 2022-04-08
:sparkles: Rework registration configuration

0x46616c6b:
I would like to simplify if the registration is open or closed. The variable `HAS_SINA_BOX` is confusing for non-german administrators/developers. Also the hardcoded limit of 9999 allowed registered users is a german special case. So my proposal is to simplify it with `REGISTRATION_OPEN`. This opens the possibility to choose it as a configuration variable without any dependency.


t2d:
You are probably right. @doobry-systemli what do you think?

t2d (approved):
Adding a comment in the Changelog would be great. Thanks!

doobry-systemli (approved):
Full ack from my side for this change :)

PR #373, opened 2022-04-12
Test also on PHP 8.1

### LINKED ISSUE #365: Support for PHP 8.0/8.1
At the moment userli is limited for PHP ^7.4. Since a while PHP 8 is out and PHP 7.4 is only in Security Support until November 28 2022. This means in December this year PHP 7.4 will not receive any updates.

See: https://www.php.net/supported-versions.php

t2d:
fixes #365


.github/workflows/integration.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.3', '7.4']
+        php-versions: ['7.3', '7.4', '8.1']
```
0x46616c6b, line 38:
```suggestion
        php-versions: ['7.4', '8.0', '8.1']
```

.github/workflows/integration.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.3', '7.4']
+        php-versions: ['7.4', '8.0', '8.1']
```
doobry-systemli, line 38: Do we want to drop PHP 7.3 support? I'd be ok with doing so and we could modernize a few things, like type declarations for class properties and use fancy new stuff like arrow functions and unpacking inside arrays.

On the other hand, this means that Userli no longer runs on Debian Buster (oldstable) systems, which is ok from my point of view.

0x46616c6b: PHP 7.3 is EOL since 2022. For me it is okay to drop.

0x46616c6b (approved):
LGTM

doobry-systemli:
Looks good to me as well. Thanks @casper <3

Issue #115, opened 2019-02-05
Two factor authentication


t2d:
Implement an easy 2FA solution with a symfony package. Maybe just OTP or U2F. It's only focused on the web interface, not IMAP/POP3.

doobry-systemli:
I started looking into this. The rough plan is:

* Second factor is only required for userli login (not IMAP/POP3/SMTP).
* For now only TOTP/google authentcator support. Webauthn etc. can come later.
* Second factor can be set, recreated and deleted in account settings. User password is required for all of them.
* Second factor gets deleted during recovery process.
* Support for backup codes.

Would be nice to hear about comments and thoughts on this :blush:

t2d:
Generally, any work on this is highly appreciated. Thanks!
Can you clarify the rationale behind just enabling it for HTTPS? In my opinion, this doesn't make too much sense without the ability to enable device-spcific passwords for IMAP etc.

t2d:
Also, have you looked at https://github.com/web-auth/webauthn-symfony-bundle?
I'm not sure, but maybe webauthn isn't much more effort than TOTP?

doobry-systemli:
By the way, I plan to use https://symfony.com/bundles/SchebTwoFactorBundle/ for the Two-factor implementation.

t2d:
I agree, thanks for explaining the reasoning.

PR #388, opened 2022-07-16
Two-factor authentication support

doobry-systemli:
Adds support for TOTP-based two-factor authentication.

## Overview

* Users can enable and disable two-factor authentication in their account settings. Both requires the user to enter their password. When two-factor authentication is enabled, the user has to verify that it's working by entering a TOTP token.
* Admins can remove the configured TOTP secret in order to restore accounts.
* The TOTP secret *doesn't* get cleared during the recovery process for security reasons. Otherwise, a single factor (recovery secret) would be enough to reset two factors (password + TOTP secret).

Fixes: #115 

### Enabling two-factor authentication in account settings

1. Settings overview
  Screenshot 2022-07-17 at 17-51-47 example org - Account Settings

2. 2FA settings, 2FA not enabled
   Screenshot 2022-07-16 at 18-34-06 example org - Two-factor authentication

3. 2FA settings, enabling 2FA
   Screenshot 2022-07-16 at 18-34-26 example org - Two-factor authentication

4. 2FA settings, enabling 2FA, invalid token
Screenshot 2022-07-16 at 18-34-47 example org - Two-factor authentication

5. 2FA settings, 2FA enabled
Screenshot 2022-07-16 at 18-35-18 example org - Two-factor authentication

### Two-factor form during login

Screenshot 2022-07-16 at 18-35-45 example org - Two-factor authentication


doobry-systemli:
Open questions where I would appreciate input:

## Do we want to reset 2FA during recovery process?

* Pro: User can restore their account when 2FA device got lost
* Con: From security perspective, recovery token is one-factor authentication. Users are asked to store it in a secure place, but still.

If we have support for 2FA backup codes, we could argue that users have to use them if they loose their 2FA device.

## Do we want admins to be able to reset 2FA for accounts?

* Pro: Accounts where 2FA device got lost can be restored by admins.
* Con: Admins can downgrade the security of accounts.

Again, with 2FA backup codes we could say that users have an option to restore 2FA themselves.

## Do we want to prod users to enable 2FA during login?

The only problem I see is that users might be overwhelmed by all the security-relevant information they need to store somewhere:
1. They have to choose a secret password and remember or store it.
2. They have to store their recovery code somewhere save.
3. They have to configure 2FA in thir 2FA app.
4. They have to store 2FA backup codes somewhere save.

That's not exactly a nice onboarding experience, is it? :thinking:

t2d:
Thanks for your awesome work. I haven't looked at the code yet, but I would like to provide some input for your questions:

> Do we want to reset 2FA during recovery process?

Yes. Recovery process is like a fresh start for your account. There'll be no more mails inside. It just makes sense to also clear the 2FA token.

> Do we want admins to be able to reset 2FA for accounts?

That's how it is commonly done and i think, it's a sensible thing to do. Especially for domain admins it'll be helpful. They can assist, but not give them selves access.

> Do we want to prod users to enable 2FA during login?

From the example I guess you're talking about signup, not login. Then my anwser: Not yet. I agree, that the onboarding experience would be degraded. Also, 2FA is not that common to regular end-users. 

In general, I'm a bit worried about confusing users with additional backup codes. But I also don't think it's possible to get rid of 2fa backup codes as you don't want to reset your whole mail account when you lose your phone. What do you think about the following idea?

When a user inserts their recovery code, give them them the choice to just delete their 2FA token or reset the whole password (and clear inbox).

doobry-systemli:
Support for backup codes is now implemented (and covered by behat tests).

Six backup codes are generated automatically during twofactor configuration. The user is asked to acknowledge that they stored the backup codes at a secure place. Only afterwards twofactor authentication is enabled.

doobry-systemli:
> Thanks for your awesome work. I haven't looked at the code yet, but I would like to provide some input for your questions:
> 
> > Do we want to reset 2FA during recovery process?
> 
> Yes. Recovery process is like a fresh start for your account. There'll be no more mails inside. It just makes sense to also clear the 2FA token.

That's wrong. Recovery process restores your account *along* with all the content.

> What do you think about the following idea?
> 
> When a user inserts their recovery code, give them them the choice to just delete their 2FA token or reset the whole password (and clear inbox).

We decided against resetting 2FA configuration with the recovery process for now. Otherwise, we would compromise the security of two-factor authentication. Being able to reset both your password and your two-factor secret using the recovery token (regardless whether it's two options in the process or one) means that *one* factor (recovery token) is enough to reset *both* factors of your account. That's not a good idea IMHO.

config/packages/scheb_2fa.yaml
```diff
@@ -0,0 +1,21 @@
...
+    backup_codes:
+        enabled: true
+
+    # The security token classes, which trigger two-factor authentication.
+    security_tokens:
+        - Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken
+        # If you're using guard-based authentication, you have to use this one:
+        # - Symfony\Component\Security\Guard\Token\PostAuthenticationGuardToken
+        # If you're using authenticator-based security (introduced in Symfony 5.1), you have to use this one:
+        # - Symfony\Component\Security\Http\Authenticator\Token\PostAuthenticationToken
```
0x46616c6b, line 21:
```suggestion
```

Can we remove these lines?

0x46616c6b (approved):

config/packages/scheb_2fa.yaml
doobry-systemli: Sure, removed :+1:

PR #402, opened 2022-10-28
Add documentation about update workflow

kidhab:
The latest Userli update requires a database schema update. Then document describes all the necessary steps.


t2d:
Thanks a lot for helping to improve userli!

PR #447, opened 2023-04-12
Remove Support for legacy Password Hashers

0x46616c6b:
We had built in a special case to cover your need to migrate old passwords to more modern hash algorithms. This is done and we can remove the special handling for old password hash algorithms.


0x46616c6b:
In future I would maybe add a new implementation but for now I would remove it, remove the complexity and reevaluate if it is necessary. We also never did another migration for quite a while now and therefor I prefer the simplicity

doobry-systemli:
Code changes look good to me, though I wonder what happens if installations still have old passwords in their database :thinking:

PR #450, opened 2023-04-17
Migrate Doctrine Entity/DB mapping to PHP annotations

doobry-systemli:
This is in preparation of migrating the mapping to PHP attributes as soon as we raise the minimum PHP version to 8.0 or higher.


t2d (approved):
I tested this branch in Vagrant and couldn't see any problems

0x46616c6b (changes requested):
I am not sure but it seems not right to put the Annotations in Traits 🤔

doobry-systemli:
> I am not sure but it seems not right to put the Annotations in Traits 🤔

Why not? Using these annotations (or later attributes) just above the variable definitions is how doctrine reads its properties. At least that's how I understood it from the documentation. And given that we use traits to define some of our properties and their getters/setters, that's the only place where the annotations can live, no?

0x46616c6b:
> > I am not sure but it seems not right to put the Annotations in Traits 🤔
> 
> Why not? Using these annotations (or later attributes) just above the variable definitions is how doctrine reads its properties. At least that's how I understood it from the documentation. And given that we use traits to define some of our properties and their getters/setters, that's the only place where the annotations can live, no?

For me, a Trait should be as simple as possible. But maybe it is a pragmatic approach, and we may not share these Traits anyway.

0x46616c6b (approved):

PR #452, opened 2023-04-20
Postpone cleaning of user object to show plainRecoveryToken before

### LINKED ISSUE #451: Automatic creation of recovery code during account creation broken
When creating a user userli just shows this empty box instead of a recovery code.

![Screenshot from 2023-04-20 16-21-48](https://user-images.githubusercontent.com/1834580/233395871-5b5ccd12-f20b-40f1-868c-7bbb913ed61e.png)

However, there's a secret box created
```
MariaDB [mail]> select id, recovery_secret_box from virtual_users where id=2;
+----+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id | recovery_secret_box                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
+----+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|  2 | GGIr0GyUn8gdROuXIwudS9DcI2oro2acpmkNpn2Olfz2djoTr5nKaUz2tgJ74osBqlh5B6U0nj7WDK6hh8K/HeXo9Qzz/Jo4F9oGCz6qx0v+wucj9/vkwgppXKp/8VA8SxNkOREIhzkWaBCVM1G+IZ0GHXqZ/1S2xraSIdBnMltyO8D7Ik4YRio0aXb9tm7c0ifPJgXei17MFmErv4UspNtphBitTkCkhaPcQBBLrOMJLKSamyVx0nU3LGgt6OV0RdH7Kln7A8tteXSLAvbyk5DBxXqak8Pmvgar7rbpLmYZy9IlTk/mwZsGWencgKQKkcChMy83p9yoexAA4hXGFi/G+jMOcvpj3+6ISE0pS7nSKTiboAdzzccdOkeY0gPc1yEjS2GZrRJZK7qY+YyZXie95XAM40OOljyX2bja2sAgq/0rUPCOr6fHgW1iQVCdB7MN/8Mk7G+XjDBQvUCtWZmILIiamrlO8BmtsBVK9+jTCWPgqWuk0h+I+god9yc2HNqDx38zDTEzp3PBTr6eJuGE0Poi+9rZO7EhVR6Jf0of8yA5LDSQKQOA1lAwuGBWzu9A3I+5Se287Xb/uBfBUnF37KYhIYWohyNvId4mFdHMzaZhVLUM48svjNPyTaUOewCu87OykB0ckM05dSRTQCHm7Idgt/+f6UwtTOtAmgu8N0ljLaZqd0N/APnn9dRvZ5qOE4s8Uee1hAm6H0vYOgjpCiFGa42YTjsQUYf2Pkq4RbwgtZ4kkw== |
+----+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

and also the UI says a recovery code is present

![image](https://user-images.githubusercontent.com/1834580/233397331-0f22876f-7f38-42dd-aec3-409f6d6e12e6.png)

Creating a new recovery code in the settings menu works.

t2d:
Fix #451

### DIFF
--- a/src/Controller/RegistrationController.php
+++ b/src/Controller/RegistrationController.php
@@ -62,7 +62,10 @@ class RegistrationController extends AbstractController
                 }
 
                 $recoveryToken = $user->getPlainRecoveryToken();
-                $user->erasePlainRecoveryToken();
+
+                // We have fetched plainRecoveryToken, which we need to show and can now remove
+                // all sensitive values from the user object
+                $user->eraseCredentials();
 
                 $recoveryTokenAck = new RecoveryTokenAck();
                 $recoveryTokenAck->setRecoveryToken($recoveryToken);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 3.2.1 (2023.04.22)
+
+* Fix display of recovery token during registration process (Fixes #451)
+* Update dependencies
+
 # 3.2.0 (2023.03.30)
 
 * Add Command to export metrics to Prometheus
--- a/features/registration.feature
+++ b/features/registration.feature
@@ -30,6 +30,7 @@ Feature: registration
 
     Then I should be on "/en/register"
     And I should see text matching "The following recovery token got created for you"
+    And I should see text matching regex "/^[0-9a-f]{8}\-[0-9a-f]{4}\-[0-9a-f]{4}\-[0-9a-f]{4}\-[0-9a-f]{12}+$/" in element with selector ".recovery-token"
 
     When I am on "/logout"
     Then I should be on "/en/"
--- a/src/Handler/RegistrationHandler.php
+++ b/src/Handler/RegistrationHandler.php
@@ -69,8 +69,8 @@ class RegistrationHandler
             $user->setMailCrypt(true);
         }
 
-        // Erase sensitive plaintext data from User object
-        $user->eraseCredentials();
+        // We used to erase sensitive data here, but it's now done in RegistrationController
+        // as we need to print the plainRecoveryToken beforehand
 
         $this->manager->persist($user);
         $this->manager->flush();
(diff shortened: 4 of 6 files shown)


src/Controller/RegistrationController.php
```diff
@@ -62,7 +62,13 @@ public function registerAction(Request $request, string $voucher = null): Respon
                 }
 
                 $recoveryToken = $user->getPlainRecoveryToken();
-                $user->erasePlainRecoveryToken();
+                if (null === $recoveryToken) {
+                    throw new \Exception('recoveryToken should not be null');
```
doobry-systemli, line 66: I don't fully get why you added this, probably a leftover from testing? What condition do you have in mind that might lead to `null === $recoveryToken` being true now that you fixed the bug? If it's merely to catch the bug, I would suggest to remove it instead.

Also you could merge the assignment and condition:

```php
if (null === $recoveryToken = $user->getPlainRecoveryToken()) {
   ...
}
```

doobry-systemli (approved):
Thanks a lot for looking into this. Your fix works well, I tested it locally.

I wonder whether it would be cleaner to keep the `$user->eraseCrendentials()` call in `RegistrationHandler->handle()` to minimize the sideeffects and things that need to be taken care of outside of the function. We could instead let `handle()` return the recovery token and pass it to the `RegistrationController->registationAction()` that way.

But searching the code it seems like we call `eraseCredentials()` all over the place, so I'm fine with your solution and everything else is probably something for a follow-up :wink:

doobry-systemli:
I added a regression test to prevent this bug in the future.

PR #507, opened 2023-11-06
Subscribe to correct LoginEvent name

doobry-systemli:
This fixes the `last_login_time` record being updated when authenticating through the `app:users:checkpassword` command.

### DIFF
--- a/tests/Command/UsersCheckPasswordCommandTest.php
+++ b/tests/Command/UsersCheckPasswordCommandTest.php
@@ -5,6 +5,8 @@ namespace App\Tests\Command;
 use App\Command\UsersCheckPasswordCommand;
 use App\Entity\User;
 use App\Enum\Roles;
+use App\Event\LoginEvent;
+use App\EventListener\LoginListener;
 use App\Handler\MailCryptKeyHandler;
 use App\Handler\UserAuthenticationHandler;
 use App\Helper\FileDescriptorReader;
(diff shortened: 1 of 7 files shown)


tests/Command/UsersCheckPasswordCommandTest.php
```diff
@@ -92,6 +102,31 @@ public function testExecuteFd3($inputStream, $returnCode): void
...
+            $reader,
+            $handler,
+            $mailCryptKeyHandler,
+            $mailCrypt,
+            $mailUID,
+            $mailGID,
+            $mailLocation);
+        $commandTester = new CommandTester($command);
+
+        $commandTester->execute([]);
```
0x46616c6b, line 127: I do not see any assertation here. It would be good to assert that the command returns successfully and the mocked services were called as expected.

mejo-: I changed the code logic now so that `$this->loginListener->expects(self::once())->method('onLogin')` is in the test function itself. Probably it's easier to follow the logic of the test that way.

I also added an assertion to make sure the `execute()` call succeeded, but that's not what we want to test here. We want to test whether `LoginListener->onLogin()` got triggered.

doobry-systemli: I changed the code logic now so that `$this->loginListener->expects(self::once())->method('onLogin')` is in the test function itself. Probably it's easier to follow the logic of the test that way.

I also added an assertion to make sure the `execute()` call succeeded, but that's not what we want to test here. We want to test whether `LoginListener->onLogin()` got triggered.

0x46616c6b (approved):

t2d (approved):
Thank you for adding the test case. Maybe you can adjust the changelog and we release a patch version afterwards?

PR #523, opened 2023-12-28
Try to reactive Dependabot for PHP Dependencies

0x46616c6b:
I was not happy to disable Dependabot for PHP at all. I understand that Symfony updates are a special case, and according to our upgrading policy, this will not work. However, for other dependencies, we can use Dependabot, and therefore, I activated it again and grouped the upgrades.

This is more or less a test and can result in failed PRs. But I am convinced that it helps to have automated updates and newer versions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,38 +1,37 @@
+---
 version: 2
 updates:
-
-  # Maintain dependencies for npm
   - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


doobry-systemli (approved):
Yay, thanks for tackling this. Full ack that we should have dependabot/renovate wherever possible :ok_hand:

PR #529, opened 2023-12-30
Disable Mosaic List in Admin

0x46616c6b:
We don't need the mosaic-style list. It does not bring any benefit.

**Before**

grafik

**After**

grafik


doobry-systemli (approved):
Agreed!

PR #531, opened 2023-12-30
Load Users with Autocomplete

0x46616c6b:
Now, not all users are fetched and put into the dropdown. You need to type something, and the users will be loaded dynamically. This significantly improves performance.

grafik

grafik


doobry-systemli (approved):
Nice, looks sensible :ok_hand: I didn't test it, just read the documentation for [`ModelAutocompleteType`](https://symfony.com/bundles/SonataAdminBundle/current/reference/form_types.html#sonata-adminbundle-form-type-modelautocompletetype)

PR #532, opened 2023-12-30
Adjust Github Actions

0x46616c6b:
- Remove `permissions` in workflows (should be the default)
- Use the fixed version runner: `ubuntu-22.04`
- Name all steps (convenience)
- Use Node LTS
- Cache Node Dependencies

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -6,8 +6,6 @@ on:
     branches:
       - main
 
-permissions: read-all
-
 env:
   DATABASE_DRIVER: pdo_sqlite
   DATABASE_URL: sqlite:///%kernel.project_dir%/var/db_test.sqlite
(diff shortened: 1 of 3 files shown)


.github/workflows/integration.yml
```diff
@@ -30,15 +28,18 @@ env:
 
 jobs:
   unit-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
doobry-systemli, line 31: What's the reason for this change? As long as there's no incompatibilities with `ubuntu-latest` I'd prefer to keep it to not have to manually bump it later. But no hard feelings here :shrug:

doobry-systemli (approved):
Good to go apart from my comment. Feel free to merge anyway, but I'd be curious about the reasoning.

.github/workflows/integration.yml
0x46616c6b: I prefer following the principle of having fixed versions instead of floating ones. This way, I aim to avoid any surprises in the future. 😄

PR #539, opened 2024-01-06
Remove extra CSS for Sonata Admin

0x46616c6b:
This is not necessary anymore. Sonata Admin Bundle comes with its own fonts and does not fetch them from Google anymore.


doobry-systemli (approved):
Nice, I trust you on that.

0x46616c6b:
> Nice, I trust you on that.

You can reproduce it in the dev environment. And we also have CSP headers in place which prevents loading fonts from remote sources

PR #548, opened 2024-01-14
Remove fallback routes

0x46616c6b:
I added a redirect from `/` to `/{_locale}` and removed the obsolete fallback routes.

- Routing is simplified and Redirect only happens on `/`. 
- PHPUnit should be installed again (Symfony Best Practice).
- Updated phpunit.xml fixes also Coverage for SonarCloud
- Removed Environment Variables from GitHub Actions (only uses `.env.test`


src/Admin/UserAdmin.php
```diff
@@ -227,8 +227,8 @@ public function setMailCryptKeyHandler(MailCryptKeyHandler $mailCryptKeyHandler)
         $this->mailCryptKeyHandler = $mailCryptKeyHandler;
     }
 
-    public function setMailCryptVar(int $mailCrypt): void
+    public function setMailCryptVar(string $mailCrypt): void
```
0x46616c6b, line 230: The parameter comes from an environment variable, so it will be a string.

composer.json
```diff
@@ -54,7 +54,10 @@
         "friends-of-behat/mink-browserkit-driver": "^1.6",
         "friends-of-behat/mink-extension": "^2.7",
         "friends-of-behat/symfony-extension": "^2.0",
+        "phpunit/phpunit": "^9.5",
```
0x46616c6b, line 57: See the configuration of testing in Symfony Docs: https://symfony.com/doc/5.x/testing.html

They recommend to install `symfony/test-pack` which also installes `phpunit/phpunit` and `symfony/css-selector`

config/packages/security.yaml
```diff
@@ -124,19 +124,16 @@ security:
...
-      - { path: "^/[a-z]{2}/2fa", roles: IS_AUTHENTICATED_2FA_IN_PROGRESS }
-      - { path: ^/admin, roles: ROLE_DOMAIN_ADMIN }
-      - { path: "^/[a-z]{2}/voucher", roles: ROLE_USER, allow_if: "!is_granted('ROLE_SUSPICIOUS')"}
-      - { path: "^/[a-z]{2}/alias", roles: ROLE_USER, allow_if: "!is_granted('ROLE_SPAM')"}
-      - { path: "^/[a-z]{2}/account", roles: ROLE_USER, allow_if: "!is_granted('ROLE_SPAM')"}
-      - { path: ^/, roles: ROLE_USER }
+      - { path: "^/$", roles: IS_AUTHENTICATED_ANONYMOUSLY }
+      - { path: "^/login", roles: IS_AUTHENTICATED_ANONYMOUSLY }
+      - { path: "^/logout", roles: IS_AUTHENTICATED_ANONYMOUSLY }
+      - { path: "^/[a-z]{2,3}/init", roles: IS_AUTHENTICATED_ANONYMOUSLY }
```
0x46616c6b, line 130: I changed the regex for locales to `[a-z]{2,3}` because we also support `gsw`, which obviously has three characters.

t2d (changes requested):
Thanks for working on this. The dependency changes in this PR are sadly not compatible with Debian 11 or 12. Can you please adjust?

0x46616c6b:
> Thanks for working on this. The dependency changes in this PR are sadly not compatible with Debian 11 or 12. Can you please adjust?

I can not reproduce this and can't install a Debian VM. Can you point me in the direction of the problem? What is the issue?

t2d (approved):
Thanks, the change looks good. The only weird thing to me is that German `de/voucher` redirects to the English `/login`. I mean the site is easy to understand, but it kind of defeats the purpose of having a multi-lingual login side

PR #556, opened 2024-01-18
Modernize Authentication and Repositories

0x46616c6b:
We implemented an `AbstractRepository` to filter out soft-deleted entities by default. This logic is also used by authentication and hides the logic a bit. In our admin interfaces, we want to see deleted entities as well, and the `AbstractRepository` was not useful for that. So the only reason I could imagine is the login flow. Therefore, Symfony has the `UserCheckerInterface` where you can decide whether a `User` can login or not.

I also removed some methods in the Repositories that are not used in any other classes.


t2d (approved):
LGTM

PR #585, opened 2024-03-28
fix: Add missing default value for roles column

0x46616c6b:
I was unable to spin up an empty database and fill it with our data fixtures. The error was:

>  An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1364 Field 'roles' doesn't have a default value


doobry-systemli (approved):
Interesting. For me it worked a few days back using `doctrine:schema:create`.

PR #588, opened 2024-03-30
feat(voucher): Check whether user is suspicious before creating voucher

doobry-systemli:


0x46616c6b (changes requested):
Can you please add a unit test for the VoucherCreateCommand?

doobry-systemli:
> Can you please add a unit test for the VoucherCreateCommand?

Yep, added now. Also commited `VoucherUserValidatorTest` which I forgot to add to the commit before.

0x46616c6b (commented):
Thanks! Can you also fix the introduced issue reported by SonarCloud?

tests/Command/VoucherCreateCommandTest.php
```diff
@@ -0,0 +1,143 @@
...
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Tester\CommandTester;
+use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Security\Core\Exception\UserNotFoundException;
+use Symfony\Component\Validator\ConstraintViolation;
+
+class VoucherCreateCommandTest extends TestCase
```
0x46616c6b, line 26: You could also utilize the KernTestCase for that: https://symfony.com/doc/current/console.html#testing-commands

doobry-systemli:
> Thanks! Can you also fix the introduced issue reported by SonarCloud?

Wasn't introduced by me, I just renamed the class. But sure, done now :blush:

tests/Command/VoucherCreateCommandTest.php
doobry-systemli: Interesting. Gave this a quick test, but it seems like this would require more changes. Since it boots the kernel and uses an actual sqlite database, we'd have to fill this database beforehands.

It would also mean a bit of a paradigm change in how we use unit tests. So far, they're stateless. With this change, unit tests would actually run integration tests. I'm not opposed to this, but so far we have behat for this.

So in any case a bigger topic.

0x46616c6b (approved):

PR #595, opened 2024-04-18
feat: Add keycloak API endpoints

doobry-systemli:
Meant to be used by our custom keycloak user provider that queries userli for users and credential validation.

## Todo

* [x] Refactor event listeners to not break stateless firewall
* [x] Tests for API endpoints (Behat or PHPUnit)

## How to test

1. Comment out `stateless: true` (currently broken) in `security.yaml`
2. Set `KEYCLOAK_API=1` in `.env.local`

List users:
`curl --header "Authorization: Bearer insecure" 'http://192.168.60.99/api/keycloak?domain=example.org&search=user'`

Get user:
`curl --header "Authorization: Bearer insecure" 'http://192.168.60.99/api/keycloak/user/user@example.org?domain=example.org'`

Validate credentials:
`curl -X POST -i --header "Authorization: Bearer insecure" --data "domain=example.org" --data "password=password" 'http://192.168.60.99/api/keycloak/validate/user@example.org'`

### DIFF
--- a/.env
+++ b/.env
@@ -54,3 +54,6 @@ MAILER_DELIVERY_ADDRESS="admin@example.org"
 # DATABASE_URL="postgresql://app:!ChangeMe!@127.0.0.1:5432/app?serverVersion=14&charset=utf8"
 DATABASE_URL="mysql://mail:password@127.0.0.1:3306/mail?charset=utf8mb4"
 ###< doctrine/doctrine-bundle ###
+
+HTTP_BASIC_AUTH_USERNAME="keycloak"
+HTTP_BASIC_AUTH_PASSWORD="keycloak"
--- a/.env
+++ b/.env
@@ -55,5 +55,7 @@ MAILER_DELIVERY_ADDRESS="admin@example.org"
 DATABASE_URL="mysql://mail:password@127.0.0.1:3306/mail?charset=utf8mb4"
 ###< doctrine/doctrine-bundle ###
 
-HTTP_BASIC_AUTH_USERNAME="keycloak"
-HTTP_BASIC_AUTH_PASSWORD="keycloak"
+### Enable keycloak API ###
+# Warning: set a secure access token and limit access to `/api/keyloak` in your webserver or reverse proxy
+KEYCLOAK_API=0
+KEYCLOAK_API_ACCESS_TOKEN="insecure"
--- a/.env
+++ b/.env
@@ -58,7 +58,7 @@ DATABASE_URL="mysql://mail:password@127.0.0.1:3306/mail?charset=utf8mb4"
 ### Enable keycloak API ###
 # Set to `true` to enable keycloak API
 KEYCLOAK_API_ENABLED=false
-# Restrict access to this IP
-KEYCLOAK_API_IP="127.0.0.1"
+# Access is restricted to these IPs (supports subnets like `10.0.0.1/24`)
+KEYCLOAK_API_IP_ALLOWLIST="127.0.0.1, ::1"
 # Warning: set a secure access token
 KEYCLOAK_API_ACCESS_TOKEN="insecure"
--- a/.env
+++ b/.env
@@ -56,6 +56,9 @@ DATABASE_URL="mysql://mail:password@127.0.0.1:3306/mail?charset=utf8mb4"
 ###< doctrine/doctrine-bundle ###
 
 ### Enable keycloak API ###
-# Warning: set a secure access token and limit access to `/api/keyloak` in your webserver or reverse proxy
-KEYCLOAK_API=0
+# Set to `true` to enable keycloak API
+KEYCLOAK_API_ENABLED=false
+# Restrict access to this IP
+KEYCLOAK_API_IP="127.0.0.1"
+# Warning: set a secure access token
 KEYCLOAK_API_ACCESS_TOKEN="insecure"
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -159,6 +159,7 @@ security:
       - { path: "^/admin", roles: ROLE_DOMAIN_ADMIN }
       - {
           path: "^/api/keycloak",
-          roles: ROLE_KEYCLOAK,
-          allow_if: "'%env(KEYCLOAK_API_ENABLED)%' == 'true' and '%env(KEYCLOAK_API_IP)%' == request.getClientIp()",
+          ips: "%env(KEYCLOAK_API_IP_ALLOWLIST)%",
+          allow_if: "'%env(KEYCLOAK_API_ENABLED)%' == 'true' and is_granted('ROLE_KEYCLOAK')",
         }
+      - { path: "^/api/keycloak", roles: ROLE_NO_ACCESS }
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -14,11 +14,11 @@ security:
         # Custom UserProvider to allow login via email and localpart (without domain)
         user:
             id: App\Security\UserProvider
-        keycloak_api_user:
+        keycloak:
             memory:
                 users:
                     - identifier: keycloak
-                      roles: ['ROLE_API_KEYCLOAK']
+                      roles: ['ROLE_KEYCLOAK']
 
     role_hierarchy:
         # User
@@ -110,10 +110,10 @@ security:
         dev:
             pattern: ^/(_(profiler|error|wdt)|css|images|js)/
             security: false
-        api_keycloak:
+        keycloak:
             pattern: ^/api/keycloak
             stateless: true
-            provider: keycloak_api_user
+            provider: keycloak
             access_token:
                 token_handler: App\Security\KeycloakApiAccessTokenHandler
         main:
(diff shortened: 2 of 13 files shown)


src/Controller/KeycloakController.php
```diff
@@ -0,0 +1,124 @@
...
+    private readonly DomainRepository $domainRepository;
+    private readonly UserRepository $userRepository;
+
+    public function __construct(private readonly EntityManagerInterface $manager, private readonly UserAuthenticationHandler $handler) {
+        $this->domainRepository = $this->manager->getRepository(Domain::class);
+        $this->userRepository = $this->manager->getRepository(User::class);
+    }
+
+    #[Route(path: '/api/keycloak', name: 'api_keycloak_index', methods: ['GET'], stateless: true)]
+    public function index(Request $request): Response
```
doobry-systemli, line 27:
```suggestion
    public function getUsersSearch(Request $request): Response
```

src/Controller/KeycloakController.php
```diff
@@ -0,0 +1,124 @@
...
+            return [
+                'id' => explode('@', $user->getEmail())[0],
+                'email' => $user->getEmail(),
+            ];
+        });
+        return $this->json($users);
+    }
+
+    #[Route(path: '/api/keycloak/count', name: 'api_keycloak_count', methods: ['GET'], stateless: true)]
+    public function count(Request $request): Response
```
doobry-systemli, line 57:
```suggestion
    public function getUsersCount(Request $request): Response
```

src/Controller/KeycloakController.php
```diff
@@ -0,0 +1,124 @@
...
+            return $this->json([
+                'message' => 'domain not found',
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        return $this->json($this->userRepository->countDomainUsers($domain));
+    }
+
+    #[Route(path: '/api/keycloak/user/{email}', name: 'api_keycloak_user', methods: ['GET'], stateless: true)]
+    public function get(Request $request, string $email): Response
```
doobry-systemli, line 69:
```suggestion
    public function getUser(Request $request, string $email): Response
```

src/Controller/KeycloakController.php
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        return $this->json([
+            'id' => explode('@', $foundUser->getEmail())[0],
+            'email' => $foundUser->getEmail(),
+        ]);
+    }
+
+    #[Route(path: '/api/keycloak/validate/{email}', name: 'api_keycloak_user_validate', methods: ['POST'], stateless: true)]
+    public function validate(Request $request, string $email): Response
```
doobry-systemli, line 94:
```suggestion
    public function postValidateUser(Request $request, string $email): Response
```

src/Enum/Roles.php
```diff
@@ -11,6 +11,7 @@ final class Roles
     public const USER = 'ROLE_USER';
     public const DOMAIN_ADMIN = 'ROLE_DOMAIN_ADMIN';
     public const ADMIN = 'ROLE_ADMIN';
+    public const API_KEYCLOAK = 'ROLE_API_KEYCLOAK';
```
doobry-systemli, line 14:
```suggestion
    public const KEYCLOAK = 'ROLE_KEYCLOAK';
```

doobry-systemli (commented):
Unit test with web test case from symfony.

src/Controller/KeycloakController.php
doobry-systemli: `getUser` is already defined by parent class `AbstractController`. Any idea for another name, @louis?

doobry-systemli: I called it `getOneUser()` now.

doobry-systemli:
I think I addressed all the remarks from review. Unit tests for `KeycloakController` are now also added.

src/Controller/KeycloakController.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        $users = $this->manager->getRepository(User::class)->findUsersByString($domainObject, $search, $max, $first)->map(function (User $user) {
+            return [
+                'id' => explode('@', $user->getEmail())[0],
+                'email' => $user->getEmail(),
+            ];
+        });
+        return $this->json($users);
+    }
```
y3n4, line 40: Could let symfonys entity resolver handle fetching the domain object or return 404 if it does not exist

```suggestion
    #[Route(path: '/api/keycloak/{domainUrl}', name: 'api_keycloak_index', methods: ['GET'], stateless: true)]
    public function getUsersSearch(
        #[MapEntity(mapping: ['domainUrl' => 'name'])] Domain $domainObject,
        #[MapQueryParameter] string $search = '',
        #[MapQueryParameter] int $max = 10,
        #[MapQueryParameter] int $first = 0,
    ): Response {
        $users = $this->manager->getRepository(User::class)->findUsersByString($domainObject, $search, $max, $first)->map(function (User $user) {
            return [
                'id' => explode('@', $user->getEmail())[0],
                'email' => $user->getEmail(),
            ];
        });
        return $this->json($users);
    }
```

.env
```diff
@@ -54,3 +54,11 @@ MAILER_DELIVERY_ADDRESS="admin@example.org"
 # DATABASE_URL="postgresql://app:!ChangeMe!@127.0.0.1:5432/app?serverVersion=14&charset=utf8"
 DATABASE_URL="mysql://mail:password@127.0.0.1:3306/mail?charset=utf8mb4"
 ###< doctrine/doctrine-bundle ###
+
+### Enable keycloak API ###
+# Set to `true` to enable keycloak API
+KEYCLOAK_API_ENABLED=false
+# Restrict access to this IP
+KEYCLOAK_API_IP="127.0.0.1"
```
0x46616c6b, line 62:
```suggestion
KEYCLOAK_API_IP_ALLOWLIST="127.0.0.1"
```
Would name that more precisely.

src/Dto/KeycloakUserValidateDto.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace App\Dto;
+
+use Symfony\Component\Validator\Constraints as Assert;
+
+class KeycloakUserValidateDto {
+    #[Assert\NotBlank]
+    public string $domain;
```
0x46616c6b, line 9: Would make the properties private and use getter and setter. We do this in other parts of the application in that way.

src/Controller/KeycloakController.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        return $this->json([
+            'id' => explode('@', $foundUser->getEmail())[0],
+            'email' => $foundUser->getEmail(),
+        ]);
+    }
+
+    #[Route(path: '/api/keycloak/validate/{email}', name: 'api_keycloak_user_validate', methods: ['POST'], stateless: true)]
+    public function postUserValidate(#[MapRequestPayload] KeycloakUserValidateDto $requestData, string $email): Response
```
0x46616c6b, line 79: Can we put the email also to the request data parameter?

0x46616c6b (changes requested):
Looks pretty good already. Have only some minor suggestions

.env
doobry-systemli: Given that it only supports one IP and no list, I think `KEYCLOAK_API_IP` is better. Or maybe `KEYCLOAK_API_ALLOW_IP`?

src/Controller/KeycloakController.php
doobry-systemli: Given that we decided to have the domain as route param, I think it's nicer to have the email as route param as well. But no strong feelings here.

.env
0x46616c6b: We must support multiple IPs or a Subnet. Otherwise it is impossible to support container based hostings or KeyCloak with HA setup.

doobry-systemli: I changed this now to use the [ips access control option](https://symfony.com/doc/6.4/security/access_control.html#matching-access-control-by-ip). This way, subnets are supported.

src/Security/KeycloakApiAccessTokenHandler.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace App\Security;
+
+use Symfony\Component\Security\Core\Exception\BadCredentialsException;
+use Symfony\Component\Security\Http\AccessToken\AccessTokenHandlerInterface;
+use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
+
+class KeycloakApiAccessTokenHandler implements AccessTokenHandlerInterface
```
0x46616c6b, line 9:
```suggestion
class KeycloakAccessTokenHandler implements AccessTokenHandlerInterface
```

config/services.yaml
```diff
@@ -164,6 +164,10 @@ services:
             - '@Doctrine\ORM\EntityManagerInterface'
         public: true
 
+    App\Security\KeycloakApiAccessTokenHandler:
```
0x46616c6b, line 167:
```suggestion
    App\Security\KeycloakAccessTokenHandler:
```

config/packages/security.yaml
```diff
@@ -104,6 +110,12 @@ security:
         dev:
             pattern: ^/(_(profiler|error|wdt)|css|images|js)/
             security: false
+        keycloak:
+            pattern: ^/api/keycloak
+            stateless: true
+            provider: keycloak
+            access_token:
+                token_handler: App\Security\KeycloakApiAccessTokenHandler
```
0x46616c6b, line 118:
```suggestion
                token_handler: App\Security\KeycloakAccessTokenHandler
```

0x46616c6b (commented):
Would suggest to shorten the TokenHandler name

src/Security/KeycloakApiAccessTokenHandler.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    public function __construct(private string $keycloakApiAccessToken)
+    {
+    }
+
+    public function getUserBadgeFrom(#[\SensitiveParameter] string $accessToken): UserBadge {
+        if (!($accessToken === $this->keycloakApiAccessToken)) {
```
0x46616c6b, line 16:
```suggestion
        if ($accessToken !== $this->keycloakApiAccessToken) {
```

src/Security/KeycloakApiAccessTokenHandler.php
doobry-systemli: Agreed and changed.

0x46616c6b (approved):

PR #598, opened 2024-04-19
Adjust Validator Signatures

0x46616c6b:
Use the correct signature from the interface:

```php
public function validate(mixed $value, Constraint $constraint): void
```

Also added tests for all validators that were missed

### DIFF
--- a/src/Validator/Constraints/EmailAddressValidator.php
+++ b/src/Validator/Constraints/EmailAddressValidator.php
@@ -6,43 +6,18 @@ use App\Entity\Alias;
 use App\Entity\Domain;
 use App\Entity\ReservedName;
 use App\Entity\User;
-use App\Repository\AliasRepository;
-use App\Repository\DomainRepository;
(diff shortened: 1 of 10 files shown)


src/Validator/Constraints/EmailAddressValidator.php
```diff
@@ -6,43 +6,18 @@
...
-use App\Repository\DomainRepository;
-use App\Repository\ReservedNameRepository;
-use App\Repository\UserRepository;
 use Doctrine\ORM\EntityManagerInterface;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\ConstraintValidator;
 use Symfony\Component\Validator\Exception\UnexpectedTypeException;
 
-/**
- * Class EmailAddressValidator.
```
mejo-, line 19: Good to say farewell to these useless class comments :wave:

mejo- (approved):

PR #601, opened 2024-04-23
PasswordChange: Use builtin password validator

y3n4:
Use Symfonys builtin validators for PasswordChange


features/user.feature
```diff
@@ -23,8 +23,8 @@ Feature: User
     And I am on "/account"
     And I fill in the following:
       | password_change_password             | asdasd       |
-      | password_change_plainPassword_first  | P4ssW0rd!!!1 |
-      | password_change_plainPassword_second | P4ssW0rd!!!1 |
+      | password_change_newPassword_first  | P4ssW0rd!!!1 |
+      | password_change_newPassword_second | P4ssW0rd!!!1 |
```
0x46616c6b, line 27:
```suggestion
      | password_change_newPassword_first    | P4ssW0rd!!!1 |
      | password_change_newPassword_second   | P4ssW0rd!!!1 |
```

src/Form/Model/PasswordChange.php
```diff
@@ -2,13 +2,37 @@
...
+    #[Assert\NotCompromisedPassword(skipOnError: 'true')]
+    #[Assert\NotIdenticalTo(propertyPath: 'password', message: 'form.identical-passwords')]
+    private string $newPassword;
+
+    public function getPassword(): ?string
+    {
+        return $this->password;
+    }
+
+    public function setPassword(?string $password): void
```
0x46616c6b, line 24:
```suggestion
    public function setPassword(string $password): void
```

I think we should be unable to `null` the password

src/Form/Model/PasswordChange.php
```diff
@@ -2,13 +2,37 @@
...
+    {
+        $this->password = $password;
+    }
+
+    public function getNewPassword(): ?string
+    {
+        return $this->newPassword;
+    }
+
+    public function setNewPassword(?string $newPassword): void
```
0x46616c6b, line 34:
```suggestion
    public function setNewPassword(string $newPassword): void
```

I think we should be unable to `null` the newPassword

0x46616c6b (changes requested):

features/user.feature
y3n4: ✅

src/Form/Model/PasswordChange.php
y3n4: ✅

0x46616c6b (approved):
Really nice improvement!

PR #602, opened 2024-04-23
Make password not optional in PasswordUpdater

0x46616c6b:
First part in the journey of removing `PlainPasswordTrait` from `User`.


doobry-systemli (approved):
Code changes look good :ok_hand:

PR #629, opened 2024-08-22
🐛 Fix MailCryptKeyHandler create/update

0x46616c6b:
Found an issue when trying to reset a user. This comes from the usage of the implicit plain password at the user object.

```
$ bin/console app:users:reset -u <REDACTED>
Really reset user? This will clear their mailbox: (yes|no) yes
New password:
Repeat password:

Resetting user <REDACTED> ...

{"message":"Error thrown while running command \"app:users:reset -u '<REDACTED>'\". Message: \"plainPassword should not be null\"","context":{"exception":{"class":"Exception","message":"plainPassword should not be null","code":0,"file":"/var/www/users.systemli.org/userli-3.8.0/src/Handler/MailCryptKeyHandler.php:77"},"command":"app:users:reset -u '<REDACTED>'","message":"plainPassword should not be null"},"level":500,"level_name":"CRITICAL","channel":"console","datetime":"2024-08-22T10:09:46.238772+00:00","extra":{}}
10:09:46 CRITICAL  [console] Error thrown while running command "app:users:reset -u '<REDACTED>'". Message: "plainPassword should not be null" ["exception" => Exception { …},"command" => "app:users:reset -u '<REDACTED>'","message" => "plainPassword should not be null"]
{"message":"Command \"app:users:reset -u '<REDACTED>'\" exited with code \"1\"","context":{"command":"app:users:reset -u '<REDACTED>'","code":1},"level":100,"level_name":"DEBUG","channel":"console","datetime":"2024-08-22T10:09:46.280854+00:00","extra":{}}

In MailCryptKeyHandler.php line 77:

  plainPassword should not be null
```

I refactored the `MailCryptKeyHandler::create` and `MailCryptKeyHandler::update` methods to use the password explictly as a parameter and not from the user itself. Also I removed the password change option in `UserAdmin`.

### DIFF
--- a/src/Handler/MailCryptKeyHandler.php
+++ b/src/Handler/MailCryptKeyHandler.php
@@ -49,10 +49,10 @@ class MailCryptKeyHandler
         sodium_memzero($privateKey);
 
         if (!$process->isSuccessful()) {
-            throw new Exception('Transforming key to PKCS#8 with OpenSSL failed. OpenSSL exited unsuccessfully: '.$process->getErrorOutput());
+            throw new Exception('Transforming key to PKCS#8 with OpenSSL failed. OpenSSL exited unsuccessfully: ' . $process->getErrorOutput());
         }
         if (!str_starts_with($process->getOutput(), '-----BEGIN PRIVATE KEY-----')) {
-            throw new Exception('Transforming key to PKCS#8 with OpenSSL failed. OpenSSL output is no valid PKCS#8 key: '.$process->getOutput());
+            throw new Exception('Transforming key to PKCS#8 with OpenSSL failed. OpenSSL output is no valid PKCS#8 key: ' . $process->getOutput());
         }
 
         return $process->getOutput();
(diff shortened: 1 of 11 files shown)


src/Handler/MailCryptKeyHandler.php
```diff
@@ -61,69 +61,51 @@ public function toPkcs8(string $privateKey): string
...
-
-        $mailCryptSecretBox = CryptoSecretHandler::create($keyPair->getPrivateKey(), $plainPassword);
+        $mailCryptSecretBox = CryptoSecretHandler::create($keyPair->getPrivateKey(), $password);
         $user->setMailCryptPublicKey($keyPair->getPublicKey());
         $user->setMailCryptSecretBox($mailCryptSecretBox->encode());
         $user->setPlainMailCryptPrivateKey($keyPair->getPrivateKey());
 
         // Clear variables with confidential content from memory
         $keyPair->erase();
-        sodium_memzero($plainPassword);
```
t2d, line 87: I don't know how to do it better, but this looks like a bit of a regression. @doobry-systemli ?

t2d (approved):
It looks consistent to me and all tests are green. So let's merge I guess

PR #635, opened 2024-08-31
♻️ Remove usage of plainPassword for User

0x46616c6b:


src/EventListener/LoginListener.php
```diff
@@ -34,25 +30,12 @@ public function onLogin(LoginEvent $event): void
     }
 
     private function handleLogin(User $user): void
-    {
-        // update password hash if necessary
```
t2d, line 38: Don't you think we'll need this functionality in the future?

0x46616c6b: No, I don't see it for the future, to be honest.

t2d: But how do you think, we'll do updates to future hashing algorithms?

0x46616c6b: We decide and implement it when it is necessary. At the moment it is dead code.

PR #641, opened 2024-09-28
✨ Add Domain Admin Settings

0x46616c6b:
Added a new tile for domain admins and a corresponding view to add two essential actions: add new accounts and add new aliases.

Bildschirmfoto am 2024-09-28 um 19 21 54

Bildschirmfoto am 2024-09-28 um 19 21 57


doobry-systemli (changes requested):
Nice work! I stumbled upon some problems though that need to be addressed.

PR #645, opened 2024-09-29
♻️ Fix adding Users in Admin

0x46616c6b:


doobry-systemli (approved):
Tested and works well. Thanks for tackling this straight away!

PR #664, opened 2024-11-29
fix(framework): Support trusted proxies for userli behind reverse proxy

doobry-systemli:


doobry-systemli:
Tested in vagrant and at least it doesn't change anything on a dev setup without a reverse proxy. No protocol is added to the generated URLs, with and without this PR :shrug:

0x46616c6b (approved):

PR #681, opened 2024-12-01
Dovecot Lua auth: Various fixes

y3n4:


contrib/checkpasswd.lua
```diff
@@ -123,14 +123,14 @@ function auth_password_verify(request, password)
...
-    http_request:set_payload(string.format('{"password": "%s"}', password))
+    http_request:set_payload(json.encode({password = password}))
     local http_response = http_request:submit()
 
     if http_response:status() == 200 then
         local data = json.decode(http_response:payload())
 
         -- mailCryptPrivateKey may be empty, but cannot be nil
-        if not(data and data.body and data.body.mailCrypt and data.body.mailCryptPrivateKey and data.body.mailCryptPublicKey) then
+        if not(data and data.body and data.body.mailCrypt and data.body.mailCryptPrivateKey and data.body.mailCryptPublicKey and data.body.quota) then
```
doobry-systemli, line 133: If I understand this line correct, the script breaks if *any* of these attributes are not set in the http response. So we only regard responses as valid that contain all the attributes we expect, right?

doobry-systemli: Ahh, now I see, we expect the attributes to be there later in the script, and probably it would fail there otherwise. So good from my side.

doobry-systemli (approved):
Code changes look good to me. I didn't test.

contrib/checkpasswd.lua
y3n4: Mostly my reasoning was that it is too complicated in lua to check for both nil and empty strings in the response body. therefore the API will always return a value, even if it is an empty string.

0x46616c6b (approved):

Issue #749, opened 2025-03-12
Accounts created via admin interface stay with disabled mail_crypt even with `MAIL_CRYPT=3`


doobry-systemli:
With `MAIL_CRYPT` >= 2, new accounts should have mail_crypt enabled per default automatically.

When we create an account via admin interface it gets a `mail_crypt_secret_box` generated automatically, but `mail_crypt` is set to `0`. Therefore, even when logging into the account with `MAIL_CRYPT=3`, `mail_crypt` doesn't get enabled later for this account.

It's probably as simple as passing `$mailCryptEnable = true` as third argument [here](https://github.com/systemli/userli/blob/0036db2c65f335fd3920f6190584c51825cc922a/src/Admin/UserAdmin.php#L180), but not sure if I miss something why this should not be done 🤔

PR #750, opened 2025-03-12
fix: Enable mail_crypt for new users via admin backend

doobry-systemli:
When environment variable `MAIL_CRYPT` is >= 2, new users should get mail_crypt enabled per default. Let's respect this also for users created via admin interface.

Fixes: #749 

This is untested so far, as I currently don't have a working development environment.

### DIFF
--- a/src/Admin/UserAdmin.php
+++ b/src/Admin/UserAdmin.php
@@ -3,6 +3,7 @@
 namespace App\Admin;
 
 use App\Entity\User;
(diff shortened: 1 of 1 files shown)


doobry-systemli:
I managed to get my dev environment working (still Vagrant :upside_down_face:) and tested both with `MAIL_CRYPT=1` and `MAIL_CRYPT=2` now.

PR #825, opened 2025-07-15
💄 Align primary button colors

0x46616c6b:
Use a consistent color (blue) for primary buttons throughout the entire project. I also adjusted the alias list after we shrank the copy button, which has now created more space.

grafik


doobry-systemli (approved):
:heart:

PR #837, opened 2025-08-04
♻️ Add JsonExceptionSubscriber

0x46616c6b:
Exceptions under `/api` or when `application/json` is expected should return JSON.

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -88,6 +88,12 @@ services:
       $defaultLocale: "%locale%"
       $supportedLocales: "%supported_locales%"
 
+  App\EventSubscriber\JsonExceptionSubscriber:
+    arguments:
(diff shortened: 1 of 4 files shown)


doobry-systemli (approved):
Code changes look good to me. Though wouldn't it be more straight forward to just catch exceptions and return `JsonResponse` in the API controllers directly instead of having this event subscriber that is hard to spot while reading the code? Not a blocker, just a question in my head.

doobry-systemli:
> Not every exception occurs inside the controller methods. Some are triggered when PropertyMappers do not find the entity, and so on. Therefore, your approach would not catch every case, which leads to mixed outputs. The listener is more generic and works like the framework. It also does not add extra code to controllers for handling exceptions in responses.

Thanks for the explanation, that makes sense.

PR #872, opened 2025-09-04
Add optional note to alias

t2d:


src/Entity/Alias.php
```diff
@@ -46,6 +46,10 @@ class Alias implements SoftDeletableInterface, Stringable
     #[ORM\Column(nullable: true)]
     protected ?string $destination = null;
 
+    #[ORM\Column(length: 255, nullable: true)]
+    #[Assert\Length(max: 255)]
```
0x46616c6b, line 50: For whatever reason, we used to do validation directly in the entity. But that's not the typical approach. Form models are designed for that. While this is already incorrect in the existing code, it should be adjusted.

src/Form/Model/AliasCreate.php
```diff
@@ -5,4 +5,5 @@
 class AliasCreate
 {
     public string $alias;
+    public ?string $note = null;
```
0x46616c6b, line 8: This is where validation should be done.

tests/Entity/AliasNoteTest.php
```diff
@@ -0,0 +1,21 @@
+<?php
```
0x46616c6b, line 1: Honestly, this test case is pretty pointless :D

src/Entity/Alias.php
0x46616c6b: And I would reduce the validation to 30-50 characters, otherwise we'll have to make adjustments to the design to accommodate a possible 255 characters.

0x46616c6b (changes requested):

PR #873, opened 2025-09-04
Allow (Domain-) Admins to delete aliases

t2d:
Admin:
image

User:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Features and Improvements
+
+- 🚸 Improve UX for deleting custom aliases (#873)
+
 ## 5.4.1 (2026.01.23)
 
 ### Bug Fixes


src/Voter/AliasVoter.php
```diff
@@ -27,18 +33,28 @@ protected function supports(string $attribute, mixed $subject): bool
...
-        $isNotDeleted = $isAliasValid && !$alias->isDeleted();
-        $isRandom = $isAliasValid && $alias->isRandom();
-        $isOwner = $isAliasValid && $alias->getUser() === $user;
-
-        return $isUserValid
-            && $isAliasValid
-            && $isNotDeleted
-            && $isRandom
-            && $isOwner;
+        $alias = $subject; // already ensured to be Alias by supports()
```
0x46616c6b, line 36: SonarCloud complains when a method has more than three return statements...

t2d:
The database doesn't know shared aliases. These are simply multiple aliases with the same source.

PR #1034, opened 2026-02-10
🚸 Improve Error Handling in Dovecot Lua Adapter

y3n4:


docker/dovecot/dovecot.conf
```diff
@@ -26,7 +26,6 @@ ssl_server_key_file = key.pem
 
 namespace inbox {
   inbox = yes
-  seperator = /
```
0x46616c6b: Why is this necessary?

PR #1081, opened 2026-02-27
🐛 Fix misleading Prometheus exporter label for user count

### LINKED ISSUE #713: Prometheus metrics report misleading user count
The Prometheus exporter calls `userRepository->countUsers()`, [which returns the __active__](https://github.com/systemli/userli/blob/f5c43cee0ca66ded940e35e192589ea2f85c268f/src/Repository/UserRepository.php#L88) user count [but falsely declares them as __total__ user count](https://github.com/systemli/userli/blob/f5c43cee0ca66ded940e35e192589ea2f85c268f/src/Command/MetricsCommand.php#L54).

y3n4:
Fix userli_users_total to actually report total user count (deleted+active users) 
Add userli_users_active_total metric

Bit unnecessary to have all three (active, deleted, total), as one could add and substract wherever one processes these metrics, but seems the best solution to preserver backwards compatibility (for example for grafana dashboards)

Closes #713

### DIFF
--- a/src/Command/MetricsCommand.php
+++ b/src/Command/MetricsCommand.php
@@ -4,12 +4,11 @@ declare(strict_types=1);
 
 namespace App\Command;
 
-use App\Entity\Alias;
-use App\Entity\Domain;
-use App\Entity\OpenPgpKey;
-use App\Entity\User;
-use App\Entity\Voucher;
-use Doctrine\ORM\EntityManagerInterface;
+use App\Repository\AliasRepository;
+use App\Repository\DomainRepository;
+use App\Repository\OpenPgpKeyRepository;
+use App\Repository\UserRepository;
+use App\Repository\VoucherRepository;
 use Override;
 use Symfony\Component\Console\Attribute\AsCommand;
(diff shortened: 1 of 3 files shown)


src/Command/MetricsCommand.php
```diff
@@ -37,12 +37,18 @@ public function __construct(private readonly EntityManagerInterface $manager)
     #[Override]
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $usersTotal = $this->manager->getRepository(User::class)->countUsers();
+        $activeUsersTotal = $this->manager->getRepository(User::class)->countUsers();
```
0x46616c6b: Since the Repositories are now `ServiceEntityRepository`, you could use this in the constructor and remove the `EntityManagerInterface`

0x46616c6b (changes requested):
An entry in UPGRADE.md would be nice to make it clear that the metrics will change. The operators should not be surprised.

0x46616c6b (approved):

PR #1171, opened 2026-03-13
Do not return random aliases in RoundcubeController

y3n4:
🚸 Exclude random aliases in Roundcube controller, up to 100 random aliases break the user experience in roundcube.


y3n4:
Will fix in roundcube as there is a usecase for random aliases as sender identity

PR #1176, opened 2026-03-14
🔒 Add domain authorization for user creation

t2d:
## Summary

Adds authorization checks to prevent domain admins from creating or editing users outside their own domain. Previously, `UserController::create` and `createSubmit` had no `denyAccessUnlessGranted` checks — a domain admin could craft a request to create users in any domain.

### The vulnerability

The `UserController::create` and `createSubmit` actions were missing authorization checks entirely. While the `DomainVoter` existed for `view`, `edit`, and `delete`, it had no `create` attribute. A domain admin with access to the admin panel could:

1. **Create users in any domain** — the form accepted any email address without verifying the domain admin had authority over that domain
2. **Escalate privileges via edit** — after form submission, a crafted POST to `editSubmit` could assign roles beyond the domain admin's authority, since the voter wasn't re-consulted after form binding

### Defense-in-depth approach

This PR implements three layers of protection, each catching what the others might miss:

1. **`DomainVoter` (authorization layer)** — Added `CREATE` attribute support. The voter denies access when a domain admin tries to create/edit a user in a domain they don't administer. This is the hard security boundary.

2. **`EmailDomainAdmin` form validator (UX layer)** — Validates the email domain *during form submission*, providing clear error messages ("You are not allowed to create users in this domain" / "The email domain does not exist") instead of a raw 403. This makes the constraint visible and understandable to the user.

3. **Controller-level `denyAccessUnlessGranted` (enforcement layer)** — Both `createSubmit` and `editSubmit` now construct an ephemeral `User` with the submitted email and roles, then call the voter. This catches edge cases where form validation alone wouldn't suffice (e.g., race conditions, direct POST manipulation).

### Key design decisions

- **`CREATE` and `DELETE` restricted to `User` subjects only** — The `AliasVoter` already handles `delete` for aliases in user-facing templates (`is_granted('delete', alias)`). Under Symfony's `unanimous` access decision strategy, if `DomainVoter` also supported `delete` on `Alias`, it would deny regular users trying to delete their own aliases (since `DomainVoter` requires `ROLE_DOMAIN_ADMIN`). Restricting these attributes to `User` avoids the collision.

- **Null domain guard** — If a domain admin somehow has no domain set, the voter denies immediately rather than risking a null comparison.

- **Non-User principal handling** — The form validator raises a violation (not a silent pass) if the authenticated principal isn't a `User` instance, preventing bypass via unexpected authentication scenarios.

- **Post-form re-authorization in `editSubmit`** — After form binding, the controller re-checks authorization with the *submitted* roles, not just the original user's roles. This prevents a domain admin from submitting a form that escalates a user to `ROLE_ADMIN`.

## Changes

- **`DomainVoter`**: Added `CREATE` constant, restricted `CREATE`/`DELETE` to `User` subjects, added null domain guard, grouped `CREATE` with `EDIT` (both use `DomainGuesser::guess()`)
- **`UserController`**: Added `denyAccessUnlessGranted` in `createSubmit` and post-form re-authorization in `editSubmit`
- **`EmailDomainAdmin` + `EmailDomainAdminValidator`**: New form-level constraint validating domain admin authority over the target email domain
- **`UserAdminType`**: Added `EmailDomainAdmin` constraint to email field (create mode only)
- **Translations**: EN and DE keys for `email-domain-not-allowed` and `email-domain-not-found`
- **Tests**: 25 tests across `DomainVoterTest` (create same/cross domain, create admin user denied, alias abstain, null domain) and `EmailDomainAdminValidatorTest` (8 tests covering all branches)

---

*Created with [opencode](https://github.com/sst/opencode) using Claude Opus 4.5*

### DIFF
--- a/src/Controller/Admin/UserController.php
+++ b/src/Controller/Admin/UserController.php
@@ -73,9 +73,7 @@ final class UserController extends AbstractController
         $form->handleRequest($request);
 
         if ($form->isSubmitted() && $form->isValid()) {
-            $user = new User($model->getEmail() ?? '');
-            $user->setRoles($model->getRoles());
-            $this->denyAccessUnlessGranted('create', $user);
+            $this->denyAccessUnlessGranted('create', $model);
 
             $this->manager->create($model);
             $this->addFlash('success', 'admin.user.create.success');
@@ -132,10 +130,9 @@ final class UserController extends AbstractController
         $form->handleRequest($request);
 
         if ($form->isSubmitted() && $form->isValid()) {
-            // Re-authorize with submitted roles to prevent escalation
-            $editedUser = new User($user->getEmail() ?? '');
-            $editedUser->setRoles($model->getRoles());
-            $this->denyAccessUnlessGranted('edit', $editedUser);
+            // Re-authorize: check submitted roles against the existing user's domain
+            $model->setEmail($user->getEmail());
+            $this->denyAccessUnlessGranted('edit', $model);
 
             $recoveryToken = $this->manager->update($user, $model);
             $this->addFlash('success', 'admin.user.edit.success');
--- a/src/Voter/DomainVoter.php
+++ b/src/Voter/DomainVoter.php
@@ -8,6 +8,7 @@ use App\Entity\Alias;
 use App\Entity\Domain;
 use App\Entity\User;
 use App\Enum\Roles;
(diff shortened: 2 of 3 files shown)


t2d:
@0x46616c6b suggestion to put on top of your change

PR #1202, opened 2026-03-17
Fix usage of AliasRepository::findByUser method

y3n4:
* ♻️ Split AliasRepository::findByUser into findByUser() (with domain filter) and findByUserAcrossDomains() (disables domain filter)
* 🐛 Fix cross domain alias search in DeleteHandler and OpenPGPKeyController:
  * Allow to to manage OpenPGP keys for cross domain aliases
  * DeleteHandler::deleteUser($user) should delete aliases of $user from all domains


t2d (approved):
LGTM

PR #1237, opened 2026-04-09
✨ Add maintenance task to remove unredeemed vouchers of deleted users

0x46616c6b:
## Summary

- Add a scheduled maintenance task that removes unredeemed vouchers belonging to soft-deleted users as a safety-net
- The task runs weekly at 05:00 via the maintenance scheduler and can be triggered manually from the admin maintenance page
- While `DeleteHandler::deleteUser()` already cleans up unredeemed vouchers synchronously during deletion, this task catches any that may have been missed (e.g. race conditions, manual DB changes)

---
<sub>The changes and the PR were generated by OpenCode.</sub>


y3n4:
Feels like a lot of new code when a database migration should also do the job, at least for existing remnants, IMO.

If we assume that race conditions can lead to a voucher not being deleted, I would rather try to fix/harden the delete handler.

Don't want to stand in your way though.

0x46616c6b:
> Feels like a lot of new code when a database migration should also do the job, at least for existing remnants, IMO.
> 
> If we assume that race conditions can lead to a voucher not being deleted, I would rather try to fix/harden the delete handler.
> 
> Don't want to stand in your way though.

It's about 70 lines of code (not including tests), which I think is okay. In comparison, a migration also requires a few lines of code and is a one-shot.

t2d (approved):
LGTM

PR #1252, opened 2026-04-18
♻️ Replace postmaster aliases with RFC address settings

0x46616c6b:
## Summary

Replaces per-domain postmaster aliases with global settings for RFC-required email addresses (`postmaster@`, `abuse@`). Instead of creating `Alias` entities for each domain, the Postfix API now resolves these addresses from configurable settings.

Bildschirmfoto am 2026-04-18 um 14 59 51

## Motivation

Previously, a `postmaster@<domain>` alias was automatically created for every new domain, pointing to `postmaster@<default-domain>`. This approach only covered `postmaster` (not `abuse` per RFC 2142), and stored routing information as database aliases that are better represented as global configuration.

## Changes

- **New settings:** `postmaster_address` and `abuse_address` in `config/settings.yaml`
- **New service:** `RfcAliasResolver` — resolves RFC addresses from settings, falls back to DB aliases
- **PostfixController:** Uses `RfcAliasResolver` instead of `AliasRepository` directly for alias lookups
- **Admin UI:** New "RFC Email Addresses" settings group with translations (EN/DE)
- **Migration:** Migrates existing postmaster alias destination to the new setting, deletes old aliases, ensures reserved names exist
- **Removed:** `CreatePostmasterAlias` message/handler and `DomainListener::onDomainCreated()` dispatch
- **Tests:** New `RfcAliasResolverTest` (5 cases), 3 new Behat scenarios for Postfix API, updated `DomainListenerTest`

---
<sub>The changes and the PR were generated by OpenCode.</sub>


t2d:
Why not create this by default?

0x46616c6b:
> Why not create this by default?

What?

t2d:
```
Found one regression to address before merging PR `#1252`:

- **Medium:** stale Postfix alias cache after RFC settings changes
- `src/Controller/Api/PostfixController.php:37` caches alias lookups (`AliasCacheKey::POSTFIX_ALIAS`) for 24h.
- This PR makes `postmaster@` / `abuse@` resolution settings-driven, but settings updates only clear `app.settings` in `src/Service/SettingsService.php:95`.
- Alias caches are not invalidated, so changing `postmaster_address` or `abuse_address` can keep returning old targets from `/api/postfix/alias/...` until TTL expiry.

t2d:
> > Why not create this by default?
> 
> What?

The behavior before was that `postmaster@` would be created automatically. You changed this to not creating these settings per default. I think this should be reversed.

PR #1254, opened 2026-04-18
🔧 Generate mail URLs from app_url setting + route path

t2d:
## Summary

The welcome, recovery-process, and alias-created mail bodies were hardcoding URL paths in every translation string via `%app_url%/<path>`. Several of those paths had drifted from the actual routes:

| Broken in translations | Actual route |
|---|---|
| `%app_url%/user/recovery_token` (en, es, nb, gsw, pt, it) | `/account/recovery-token` |
| `%app_url%/recovery_token` (fr typo) | `/account/recovery-token` |
| `%app_url%/voucher` (en, de, es, fr, nb, pt, it) | `/account/voucher` |
| `%app_url%/alias` (en, de, es, fr, gsw, pt, it) | `/account/alias` |
| `%app_url%/recupero` (it typo) | `/recovery` |

Rather than patch each language, the three mailers now inject `UrlGeneratorInterface` and pass the URLs as dedicated placeholders (`%recovery_url%`, `%recovery_token_url%`, `%voucher_url%`, `%alias_url%`). Route renames propagate automatically; translators no longer carry URL fragments as copy.

## URL construction

Mails are built in messenger background workers where no `Request` is on the stack, so `ABSOLUTE_URL` generation (which needs a host) is not an option. Hybrid approach:

- **Host** comes from `SettingsService->get('app_url')` — admin-editable, unchanged source of truth.
- **Path** comes from `UrlGenerator::generate(\$route, [], ABSOLUTE_PATH)` — no Request context required.

## Changes

- `src/Mail/{RecoveryProcessMailer,WelcomeMailer,AliasCreatedMailer}.php` — inject `UrlGeneratorInterface`, expose `%recovery_url%`, `%recovery_token_url%`, `%voucher_url%`, `%alias_url%`; drop the hardcoded `%app_url%/…` fragments.
- `default_translations/{en,de,es,fr,gsw,it,nb,pt}/messages.*.yml` — switch to the new placeholders (also fixes the `/recupero` and `/recovery_token` typos).
- `tests/Mail/*MailerTest.php` — unit tests updated with `UrlGeneratorInterface` stubs asserting the new placeholders and their values.
- `features/mail_links.feature` + `tests/Behat/MailCollector.php` + `tests/Behat/FeatureContext.php` — new Behat scenario drives the real `POST /recovery` flow with an admin-set `app_url`, captures the dispatched `Email` via a `MessageEvent` subscriber, and asserts the generated mail body contains the expected absolute URLs.
- `config/packages/mailer.yaml` — under `when@test` only, set `message_bus: false` so mails dispatch synchronously through the null transport (the in-memory async transport would queue them forever in tests and `MessageEvent` would never fire). Production behaviour is untouched — `APP_ENV=prod` still routes through `messenger.bus.default` / the async transport.
- `src/AGENTS.md` — new "Mail URLs" section plus anti-pattern line so the next contributor reaches for a URL placeholder instead of hardcoding paths into translations.

Non-mail uses of `%app_url%` (registration welcome page links to `/account/twofactor` and `/account/openpgp`) are intentionally left alone — those paths are correct and the pattern there is Twig-side `|trans`, not mail copy.

## Test plan

- [x] \`./bin/phpunit --filter \"RecoveryProcessMailerTest|WelcomeMailerTest|AliasCreatedMailerTest\"\` — all green
- [x] \`./bin/behat features/mail_links.feature\` — all green
- [ ] Trigger the flows manually via fixtures (\`make fixtures\`), open mailcatcher, verify every link in the welcome / recovery / alias mail resolves to a real route
- [ ] Verify Weblate picks up the changed source strings (expected — English/German source edit; other languages will be marked for re-translation)


t2d:
@0x46616c6b i think this will work. Which tests can be removed in you opinion?

0x46616c6b (approved):

t2d:
finally! 😆

PR #1319, opened 2026-07-13
🐛 fix: Only resolve RFC aliases for managed domains

0x46616c6b:
## Summary

Fixes a bug introduced in #1252: the Postfix alias endpoint resolved `postmaster@` and `abuse@` addresses from the RFC address settings for **any** domain — including domains not managed by this instance. When Postfix queried an alias like `postmaster@unmanaged.example`, the API returned the configured RFC destination, making the instance appear responsible for mail it should not handle.

## Changes

- `RfcAliasResolver::isRfcAddress()` now also verifies that the domain part of the source address is a managed domain (via `DomainRepository::existsByName()`). Addresses on unmanaged domains fall back to the regular alias lookup, which yields an empty result.
- Unit tests extended with cases for unmanaged domains.
- New Behat scenario asserting that `postmaster@unmanaged.org` returns `[]` even when `postmaster_address` is configured.

### DIFF
--- a/features/api_postfix.feature
+++ b/features/api_postfix.feature
@@ -73,6 +73,19 @@ Feature: Postfix API
             """
 
     @alias
+    Scenario: Get postmaster alias for unmanaged domain returns empty
+        Given the following Setting exists:
+            | name                | value              |
+            | postmaster_address  | admin@example.org  |
+        And I have a valid API token "postfix-test-123"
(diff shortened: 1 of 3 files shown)


doobry-systemli (approved):
Code changes look sensible to me. Thanks for fixing this ❤️

PR #1321, opened 2026-07-13
⬆️ Upgrade to TypeScript 7 with Babel-based transpilation

0x46616c6b:
## Summary

Dependabot's attempt to bump TypeScript to 7.x in #1317 fails because TypeScript 7 is the new native Go compiler, which no longer exposes the JavaScript compiler API that `ts-loader` depends on (`TypeError: Cannot read properties of undefined (reading 'fileExists')`).

This PR makes the project TypeScript 7 ready instead of pinning the dependency:

- Switch Webpack Encore from `.enableTypeScriptLoader()` (ts-loader) to `.enableBabelTypeScriptPreset()`. Babel only strips types during bundling and does not need the TypeScript compiler API. The tsconfig already sets `isolatedModules: true`, which this transpilation mode requires.
- Upgrade `typescript` to `^7.0.2`, add `@babel/preset-typescript`, remove `ts-loader`.
- Since Babel does not type-check, add a dedicated `yarn typecheck` script (`tsc --noEmit`) and run it in the integration workflow between the asset build and Vitest, so type checking is preserved as a CI gate.

As a side effect, both the typecheck (native compiler) and the Webpack build (no compiler API round-trips) get faster.

Closes #1317 (the postcss bump from that group PR will come through cleanly in the next Dependabot run).

### DIFF
--- a/src/DataFixtures/LoadReservedNameData.php
+++ b/src/DataFixtures/LoadReservedNameData.php
@@ -15,6 +15,7 @@ final class LoadReservedNameData extends Fixture implements FixtureGroupInterfac
     private const array RESERVED_NAMES = [
         'admin',
         'root',
+        'abuse',
         'webmaster',
     ];
 
--- a/src/DataFixtures/LoadReservedNameData.php
+++ b/src/DataFixtures/LoadReservedNameData.php
@@ -15,7 +15,6 @@ final class LoadReservedNameData extends Fixture implements FixtureGroupInterfac
     private const array RESERVED_NAMES = [
         'admin',
         'root',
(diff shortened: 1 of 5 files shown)


src/DataFixtures/LoadReservedNameData.php
```diff
@@ -15,6 +15,7 @@ final class LoadReservedNameData extends Fixture implements FixtureGroupInterfac
     private const array RESERVED_NAMES = [
         'admin',
         'root',
+        'abuse',
```
0x46616c6b:
```suggestion
```

